PR #7, opened 2021-08-19
Added scenarios and step definitions for the feature file

karunak21:
AS-2552
Added Scenarios and step definitions to the feature file.


e2e-tests/package.json
```diff
@@ -25,6 +25,7 @@
   },
   "devDependencies": {
     "axe-core": "^4.3.2",
+    "cucumber": "^7.0.0-rc.0",
```
shilpabasu22, line 28: we should be using cypress-cucumber-preprocessor

shilpabasu22 (approved):
LGTM

karunak21 (commented):
All comments have been addressed

PR #14, opened 2021-09-14
Feature/as 2855 description page

shepda:
Added a new page for Valid Appeal Details, which is displayed to the user if they select 'Valid' in the Review Appeal Submission page. Either of the two other options on the Review Appeal Submission page take the user to the Appeals List for now until those pages are built.

Also the following
 
- added and updated common view helpers, which includes switching the form component to a form template and added button and button group (which includes a link) components
- added mondodb and session config so we can store data in the session
- added skeleton functions to get and save data, which can be called from the controllers of new pages to hopefully help decouple these from needing to interact with the database. These can be connected to the API in a future ticket to actually get and save the data.
- moved the test mocks and added some new ones
- added a saveAndContinue common function which can be called from the POST action of the page controllers and should remove some of the otherwise duplicate code needed to handle field validation, displaying an error and redirecting to the next page
- updated existing files to to use the new common utils, such as test mocks, saveAndContinue function in the controller POST
    action and form template

This PR has been split into smaller commits to hopefully make reviewing a bit easier.


packages/web-app/src/lib/get-case-data.js
```diff
@@ -0,0 +1,22 @@
...
+  const {
+    log,
+    params: { appealId },
+    session,
+  } = req;
+
+  try {
+    if (session && !session.appeal) {
+      log.debug({ id: appealId }, 'Getting existing data');
+      session.appeal = getData(appealId);
```
swdpcomputing, line 13: Bit confused by this. It looks like this is populating mock data, but shouldn't it populate something a bit more generic?

shepda: It's populating mock data for now until we can connect the APIs to the database then it can get the submission data for the appeals/replies as necessary

packages/web-app/src/validation/valid-appeal-details.spec.js
```diff
@@ -0,0 +1,61 @@
...
+      {
+        value: undefined,
+        msg: 'Enter a description of development',
+        param: 'valid-appeal-details',
+        location: 'body',
+      },
+    ]);
+  });
+
+  it('should fail validation when given an empty string', async () => {
```
swdpcomputing, line 41: Is this to test if something is written, then deleted?

packages/web-app/src/controllers/valid-appeal-details.js
```diff
@@ -0,0 +1,52 @@
...
+      appeal: {
+        appeal: { id, horizonId },
+      },
+    },
+    body,
+  } = req;
+
+  req.session.appeal.casework.validAppealDetails = req.body['valid-appeal-details'];
+
+  saveAndContinue({
```
swdpcomputing, line 40: Is seperating the save and continue functionality to declutter controller, or is it for reuseablity?

swdpcomputing:
When finishing the process on one appeal submission, the remaining text in the Description of development box remains rather than being blanked.

packages/web-app/src/validation/valid-appeal-details.spec.js
shepda: Sure, I've removed the test

PR #15, opened 2021-09-16
feat(document download from document-service): document download

biroldif:


docker-compose.yml
```diff
@@ -31,8 +34,9 @@ services:
...
-    container_name: mongo-db
-    image: mongo:3.6.21
-    ports:
-      - 4000:27017
+    networks:
+      - appeal-planning-decision_default
+
+networks:
+  appeal-planning-decision_default:
+    external: true
```
shepda, line 42: By linking this file by the network with the docker-compose file from the Appeals Service, this presumably means that it has a dependency on the Appeals Service file and both need to at the same time.

Perhaps this is inevitable as we continue to develop both apps but as they are currently separate entities in different repos it not necessarily immediately obvious that both need to be running.

Perhaps we need to have a wider discussion about how we best handle that, for example to we merge the docker-compose files and have a script that pulls both repos from GitHub, or something similar? 

I think at the least we should have some documentation in the README of this repo to detail this in terms of how to run the app.

@rcmsoftwaredevelopment @AnthonyJWalsh Any thoughts?

packages/web-app/src/controllers/document-service-proxy.js
```diff
@@ -0,0 +1,34 @@
...
+      'content-disposition': `attachment;filename="${fileName}"`,
+      'content-type': fileResponse.headers.get('content-type'),
+    });
+  } catch (e) {
+    logger.error({ e }, 'Document was retrieved but appears to be missing key data.');
+    return res.sendStatus(500);
+  }
+
+  return fileResponse.body.pipe(res);
+};
```
shepda, line 34: It doesn't appear that we have any unit tests for `controllers/document-service-proxy.js`?

packages/web-app/src/lib/documents-api-wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+      })
+    );
+  } catch (err) {
+    logger.error({ err }, 'Error');
+    throw err;
+  }
+}
+
+exports.getDocument = async (appealId, documentId) =>
+  handler(`/api/v1/${appealId}/${documentId}/file`);
```
shepda, line 61: It doesn't appear that we have any unit tests for `lib/documents-api-wrapper.js`?

packages/web-app/src/routes/document-service-proxy.js
```diff
@@ -0,0 +1,9 @@
+const express = require('express');
+
+const router = express.Router();
+
+const documentServiceProxyController = require('../controllers/document-service-proxy');
+
+router.get('/:appealId/:documentId', documentServiceProxyController.getDocument);
+
+module.exports = router;
```
shepda, line 9: It doesn't appear that we have any unit tests for `routes/document-service-proxy.js`?

packages/web-app/src/views/components/unorderedList.njk
```diff
@@ -1,9 +1,9 @@
-{% macro unorderedList(items) %}
+{% macro unorderedList(items, rootUrl = '') %}
   <ul class="govuk-list">
 
     {% for item in items %}
-      <li><a class="govuk-link" href="#">{{ item.name }}</a></li>
+      <li><a class="govuk-link" href="{{ rootUrl }}/{{ item.id }}">{{ item.name }}</a></li>
```
shepda, line 5: I can see why you've added  `item.id` to the end of the `href` but I have a slight concern that this could cause us an issue in the future if we want to use this component but don't want the item id added to the end of the URL.

The only other way I can think of doing it is to manipulate the data in the `review-appeal-submission` template before its passed into the component.

We can leave it for now, just something to be aware of.

packages/web-app/src/lib/documents-api-wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+  const correlationId = uuid.v4();
+  const url = `${config.documents.url}${path}`;
+
+  const logger = parentLogger.child({
+    correlationId,
+    service: 'Documents Service API',
+  });
+
+  try {
+    logger.debug({ url, method, opts, headers }, 'New call');
```
shepda, line 18: I think it would make the logging more useful if the log message included the current action so that it's clearer where the log message originates from in the code.

For example `Fetching document from Blob Storage`

packages/web-app/src/lib/documents-api-wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+          method,
+          headers: {
+            'X-Correlation-ID': correlationId,
+            ...headers,
+          },
+          ...opts,
+        });
+
+        if (!apiResponse.ok) {
+          logger.debug(apiResponse, 'API Response not OK');
```
shepda, line 33: I think it would make the logging more useful if the log message included the current action so that it's clearer where the log message originates from in the code.

For example `Failed to fetch document from Blob Storage`

packages/web-app/src/lib/documents-api-wrapper.js
```diff
@@ -0,0 +1,61 @@
...
+
+            /* istanbul ignore next */
+            throw new Error(apiResponse.statusText);
+          } catch (e) {
+            logger.error({ e }, 'Error on documents API wrapper handler.');
+            throw new Error(e.message);
+          }
+        }
+
+        logger.debug('Successfully called');
```
shepda, line 49: I think it would make the logging more useful if the log message included the current action so that it's clearer where the log message originates from in the code.

For example `Successfully fetched document from Blob Storage`

shepda (commented):
Please have a look at the unit tests as some of the existing ones are failing. 

Are there any security concerns that we need to consider regarding who can access the documents? @rcmsoftwaredevelopment @AnthonyJWalsh Any thoughts?

docker-compose.yml
AnthonyJWalsh: I think this dependency is probably acceptable. Realistically our only options to mitigate this would be to set up a mock document service within the back-office repo (time, doesnt give us anything really helpful), or to go back to a mono-repo (has its own problems). Happy to chat about it though.

packages/web-app/src/lib/documents-api-wrapper.js
biroldif: document-service-proxy.spec.js added

packages/web-app/src/views/components/unorderedList.njk
biroldif: document-service-proxy.spec.js added

packages/web-app/src/controllers/document-service-proxy.js
biroldif: document-service-proxy.spec.js added

packages/web-app/src/lib/documents-api-wrapper.js
biroldif: fixed

packages/web-app/src/lib/documents-api-wrapper.js
biroldif: fixed

packages/web-app/src/lib/documents-api-wrapper.js
biroldif: fixed

shepda: There is a typo in the new error message, `successfull` should be `successful` :)

biroldif: should be fine now :)

shepda: :+1:

packages/web-app/src/lib/api-wrapper.js
```diff
@@ -2,7 +2,7 @@ const getData = (appealId) => {
   if (appealId) {
     return {
       appeal: {
-        id: '1dc8a211-4e1b-44fc-9e9c-a6e7b802fb8a',
+        id: appealId || '1dc8a211-4e1b-44fc-9e9c-a6e7b802fb8a',
```
shepda, line 5: The addition of this condition for the scenario where `appealId` is not set and the default value is used is not covered in the tests, please add a test to cover this.

shepda (approved):
LGTM

PR #21, opened 2021-09-28
feat(web-app): create questionnaire status page

swdpcomputing:


packages/web-app/src/controllers/questionnaires-list.js
```diff
@@ -0,0 +1,84 @@
...
+    const oNew = { ...o };
+    oNew.status.colour =
+      // eslint-disable-next-line no-nested-ternary
+      oNew.status.text === 'Overdue' ? 'red' : oNew.status.text === 'Received' ? 'green' : 'blue';
+    return oNew;
+  });
+
+const questionnairesList = (req, res) => {
+  const colourisedStatuses = colouriseStatuses(getRawData());
+  const sortedQuestionnairesData = sortBy(colourisedStatuses, (o) => getTime(new Date(o.dueDate)));
```
shepda, line 76: Do we need to import lodash to use the `sortBy` function or is this something we could do with the built in `sort` function?

packages/web-app/src/controllers/questionnaires-list.js
```diff
@@ -0,0 +1,84 @@
...
+    status: { text: 'Overdue' },
+  },
+];
+
+const colouriseStatuses = (data) =>
+  data.map((o) => {
+    const oNew = { ...o };
+    oNew.status.colour =
+      // eslint-disable-next-line no-nested-ternary
+      oNew.status.text === 'Overdue' ? 'red' : oNew.status.text === 'Received' ? 'green' : 'blue';
```
shepda, line 70: Perhaps rather than disabling the eslint check here we could use a status to colour mapping function?

This would make the code more easily extensible should we need to add a new status to colour map in the future.

packages/web-app/src/views/questionnaires-list.njk
```diff
@@ -0,0 +1,37 @@
...
+
+{% for questionnaire in sortedQuestionnairesData %}
+  {% set status %}
+      <strong class="govuk-tag govuk-tag--{{ questionnaire.status.colour }}">
+        {{ questionnaire.status.text }}
+      </strong>
+  {% endset %}
+  
+  {% set questionnairesListDisplayData = (questionnairesListDisplayData.push([
+    { html: "<a href=\"review-questionnaire-submission/" + questionnaire.id + "\">" + questionnaire.reference + "</a>" },
```
shepda, line 15: Could we please add the `govuk-link` class to the link?

We could also consider adding the `govuk-link--no-visited-state` class as it's a caseworking system the visited state may provide no benefit to the user and potentially adds distracting colour? Perhaps a question for the design team.

packages/web-app/src/controllers/questionnaires-list.js
```diff
@@ -0,0 +1,84 @@
...
+    return oNew;
+  });
+
+const questionnairesList = (req, res) => {
```
shepda, line 74: We generally prefix the action with the http verb so for consistency perhaps we'd be better to rename this to `getQuestionnairesList`?

packages/web-app/src/views/questionnaires-list.njk
```diff
@@ -0,0 +1,37 @@
...
+    { text: questionnaire.site },
+    { html: status, classes: 'govuk-table__cell govuk-table__cell--numeric' }
+  ]), questionnairesListDisplayData) %}
+{% endfor %}
+
+{% block backButton %}
+{% endblock %}
+
+{% block content %}
+  <h1 class="govuk-heading-xl">{{ pageTitle }}</h1>
```
shepda, line 26: We're missing the breadcrumb navigation from just above the page title, is this something that we need to include on this page? We don't have any breadcrumb navigation on the Appeals List page so should we keep these pages consistent or is that not necessary? Perhaps a question for the design team.

swdpcomputing: The ticket states that this should not involve the breadcrumbs.

packages/web-app/src/controllers/questionnaires-list.js
swdpcomputing: I think Lodash means a more terse line of code, and probably provides faster, better tested sorting for what could be large objects.

packages/web-app/src/controllers/questionnaires-list.js
swdpcomputing: I was umming and ahhing over this. Changed.

packages/web-app/src/views/questionnaires-list.njk
swdpcomputing: Done.

packages/web-app/src/controllers/questionnaires-list.js
swdpcomputing: Done.

packages/web-app/src/controllers/questionnaires-list.js
shepda: I don't think we have any benchmarks on how the two functions perform with regards to speed or scalability and I guess this is more down to personal preference but I'd almost always favour using the built in methods rather than using from an external library.

packages/web-app/src/views/questionnaires-list.njk
shepda: The `govuk-link` class should be on the `a` element rather than on the parent `td` element.

packages/web-app/src/views/questionnaires-list.njk
shepda: So it does, missed that the first time round :)

packages/web-app/src/views/questionnaires-list.njk
shepda: The class still appears to be on the `td` element. The class needs to be added to the link on line 9, something like 

```
{% set reference = "<a href=\"review-questionnaire-submission/" + questionnaire.id + "\" class=\"govuk-link\">" + questionnaire.reference + "</a>" %}
```

swdpcomputing: Done.

packages/web-app/src/views/questionnaires-list.njk
```diff
@@ -0,0 +1,43 @@
...
+  {% endif %}
+
+  {% set status %}
+    <strong class="govuk-tag govuk-tag--{{ questionnaire.status.colour }}">
+      {{ questionnaire.status.text }}
+    </strong>
+  {% endset %}
+  
+  {% set questionnairesListDisplayData = (questionnairesListDisplayData.push([
+    { html: reference, classes: referenceClass },
```
shepda, line 21: I think we could remove `classes: referenceClass` as `referenceClass` is no longer being set?

shepda (approved):
LGTM

PR #22, opened 2021-09-30
feat(missing or wrong review): page implementation

biroldif:


packages/web-app/src/views/missing-or-wrong.njk
```diff
@@ -0,0 +1,126 @@
...
+          html: otherHtml
+        }
+      }
+    ]
+  }) }}
+
+  {{ buttonGroup(backLink) }}
+{% endblock %}
+
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
```
shepda, line 126: It doesn't look like we need to import `govukTextarea` here as it's not being used in the template.

shepda (approved):
LGTM

PR #24, opened 2021-10-04
test: added tests for the case officer page

shilpabasu22:


rcmsoftwaredevelopment (approved):

shepda:
The commit message is in the wrong format which will probably cause issues with the commit message linting so please edit it before merging to master.

The commit message should be something like `commit-type(ticket-number): the commit message` and can be easily created by running `npm run commit` from the root of the repo.

PR #26, opened 2021-10-08
feat(checkandconfirm): page implementation to support valid, invalid, incomplete review outcomes

biroldif:


packages/web-app/src/routes/index.js
```diff
@@ -26,5 +27,6 @@ router.use(`/${views.invalidAppealDetails}`, handleAppealAlreadyReviewed, invali
 router.use(`/${views.missingOrWrong}`, handleAppealAlreadyReviewed, missingOrWrongAppealDetails);
 router.use(`/${views.document}`, documentsServiceProxy);
 router.use(`/${views.appealAlreadyReviewed}`, appealAlreadyReviewed);
+router.use(`/${views.checkAndConfirm}`, checkAndConfirmDetails);
```
demzzydev, line 30: might want to consider using `submissions-for-review/check-and-confirm` instead so that the full route is captured since I have a ticket that is very similar to this.

packages/web-app/src/views/missing-or-wrong.njk
```diff
@@ -120,5 +120,5 @@
     ]
   }) }}
 
-  {{ buttonGroup(backLink) }}
+  {{ buttonGroup('/check-and-confirm') }}
```
shepda, line 123: The 'Change outcome' link should be set to `backLink` as it wants to return to the previous page rather than '`check-and-confirm`, which will go to the next page?

packages/web-app/src/views/check-and-confirm.njk
```diff
@@ -0,0 +1,117 @@
+{% extends "layouts/form.njk" %}
+
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
```
shepda, line 4: It looks like `govukRadios` is not being used, do we need to import it?

packages/web-app/src/views/check-and-confirm.njk
```diff
@@ -0,0 +1,117 @@
...
+{% from "components/address.njk" import address %}
+{% from "components/unorderedList.njk" import unorderedList %}
```
shepda, line 6: It seems that `unorderedList` is not being used, do we need to import it?

packages/web-app/src/views/check-and-confirm.njk
```diff
@@ -0,0 +1,117 @@
...
+{% set originalApplicationFile = appealData.requiredDocumentsSection.originalApplication.uploadedFile %}
+{% set decisionLetterFile = appealData.requiredDocumentsSection.decisionLetter.uploadedFile %}
+{% set appealStatementFile = appealData.yourAppealSection.appealStatement.uploadedFile %}
+{% set outcomeDetails = reviewOutcome.outcomeDetails %}
+{% set invalidAppealDetails = reviewOutcome.invalidAppealDetails %}
+{% set validAppealDetails = reviewOutcome.validAppealDetails %}
+{% set completedError = errors['check-and-confirm-completed'].msg %}
+
+{% macro renderReason(reason) %}
+  <p class="govuk-body govuk-!-margin-bottom-2">{{ reason }}</p>
```
shepda, line 21: I don't think we need to put this text in a `p` tag because it's part of the `dl` created by `govukSummaryList`. The other `dd` values by default aren't wrapped in a `p` tag so not using one here would keep these consistent and keep the margin bottom the same for each, which I see that you've tried to correct with `govuk-!-margin-bottom-2` but it is still slightly larger than the other values.

PR #28, opened 2021-10-12
feat(as-3509): update azure pipeline config

shepda:
Add Dockerfile, Azure pipeline config and update npm scripts for both the web-app and api.

Also moved some functions from the Back Office Common package into the web-app because, although this works fine locally, in Azure the Back Office uses the Appeals Service Common package and this version of the Common package doesn't have these functions, hence causing unit test failures. We could have moved them to this Common package but they are only used in the Back Office web-app and we're in the process of moving the Common package to the common repo.


packages/api/azure-api-build.yml
```diff
@@ -0,0 +1,148 @@
+# Docker
+# Build and push an image to Azure Container Registry
+# https://docs.microsoft.com/azure/devops/pipelines/languages/docker
+
+trigger:
+  branches:
+    include:
+      - feature/*
```
demzzydev, line 8: We don't want this to build on the feature branches either

packages/web-app/azure-web-app-build.yml
```diff
@@ -0,0 +1,148 @@
+# Docker
+# Build and push an image to Azure Container Registry
+# https://docs.microsoft.com/azure/devops/pipelines/languages/docker
+
+trigger:
+  branches:
+    include:
+      - feature/*
```
demzzydev, line 8: We don't want this to build on the feature branches either

packages/web-app/src/lib/to-array.js
```diff
@@ -0,0 +1,13 @@
+const toArray = (candidate) => {
```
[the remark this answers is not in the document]
shepda: See my comment in the PR description.

rcmsoftwaredevelopment (approved):

PR #30, opened 2021-10-13
Questionnaire Check & Confirm

demzzydev:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-3496

## Description of change
Adding questionnaire check & confirm page for complete & incomplete journeys.

## Checklist

- [ ] Requires infrastructure changes
- [ ] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [ ] I have updated the documentation accordingly
- [ ] My commit history in this PR is linear
- [x] New features have tests
- [ ] Breaking change (team conversation required)

## Important

Please do not merge from `master` (please only [rebase](https://github.com/foundry4/appeal-planning-decision/wiki/An-intro-to-Git-Rebase)). This keeps the history linear and easier to debug.

### DIFF
--- a/packages/web-app/src/controllers/questionnaire-check-and-confirm.js
+++ b/packages/web-app/src/controllers/questionnaire-check-and-confirm.js
@@ -130,12 +130,11 @@ const getBreadcrumbs = () => [
 ];
 
 const getCheckAndConfirm = (req, res) => {
-  const { appealId } = req.param;
   const { outcome } = req.session;
 
   res.render('questionnaire-check-and-confirm', {
     pageTitle: 'Review questionnaire',
-    sections: getConfirmationSections(appealId, outcome),
+    sections: getConfirmationSections(outcome),
     breadcrumbs: getBreadcrumbs(),
     reviewOutcome: outcome,
   });


packages/web-app/src/controllers/questionnaires-for-review/check-and-confirm.js
```diff
@@ -0,0 +1,76 @@
+const {
+  questionnairesForReview: previousPage,
+  checkAndConfirm: currentPage,
+  home: nextPage,
```
shepda, line 4: It doesn't look like `nextPage` is being used so do we need to import it here?

packages/web-app/src/controllers/questionnaires-for-review/check-and-confirm.js
```diff
@@ -0,0 +1,76 @@
...
+    title: 'Appeal site',
+    value: 'Site',
+  },
+  {
+    title: 'Local planning department',
+    value: 'Newark and Sherwood District Council',
+  },
+];
+
+const getConfirmationSections = async (appealReference, outcome) => {
```
shepda, line 31: It doesn't look like this function needs to be `async`

packages/web-app/src/controllers/questionnaires-for-review/check-and-confirm.js
```diff
@@ -0,0 +1,76 @@
...
+      });
+      return sections;
+    }
+
+    default:
+      return sections;
+  }
+};
+
+const getBreadcrumbs = async () => [
```
shepda, line 46: It doesn't look like this function needs to be `async`

packages/web-app/src/controllers/questionnaires-for-review/check-and-confirm.spec.js
```diff
@@ -0,0 +1,43 @@
+const getCheckAndConfirmController = require('./check-and-confirm');
+const views = require('../../config/views');
+
+describe('controllers/check-and-confirm', () => {
+  describe('getCheckAndConfirm', () => {
+    it('should render view with complete status', async () => {
```
shepda, line 6: Could we also add a test for an Incomplete status?

packages/web-app/src/routes/questionnaires-for-review/check-and-confirm.spec.js
```diff
@@ -0,0 +1,17 @@
+const { get } = require('../../../test/routes/router-mock');
+const checkAndConfirmController = require('../../controllers/questionnaires-for-review/check-and-confirm');
+
+describe('routes/check-and-confirm', () => {
```
shepda, line 4: Could the description use the full path, for example `routes/questionnaires-for-review/check-and-confirm`?

packages/web-app/src/routes/questionnaires-for-review/check-and-confirm.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    // eslint-disable-next-line global-require
+    require('./check-and-confirm');
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('should define the expected routes', () => {
+    expect(get).toHaveBeenCalledWith('/:appealId', checkAndConfirmController.getCheckAndConfirm);
```
shepda, line 15: It looks like the path should be `/` as the route's not using the `appealId` param.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
```diff
@@ -0,0 +1,33 @@
...
+            {{ confirmlist(sections) }}
+        </div>
+    </div>
+    {% if reviewOutcome === 'COMPLETE' %}
+        <h3 class="govuk-heading-m">Confirming the questionnaire as complete makes the appeal available to inspectors </h3>
+    {% endif %}
+    <div class="govuk-button-group">
+        <a href="confirm" role="button" draggable="false" class="govuk-button"> Confirm outcome </a>
+        <a class="govuk-link" href="review">Go back to review questionnaire</a>
+    </div>
```
shepda, line 32: Could the `buttonGroup` component be used here instead of adding the HTML?

packages/web-app/src/views/questionnaire-check-and-confirm.njk
```diff
@@ -0,0 +1,33 @@
...
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-xl">
+                Review questionnaire
+            </h1>
+            <h2 class="govuk-heading-l">Check and confirm</h2>
+            {{ confirmlist(sections) }}
+        </div>
+    </div>
+    {% if reviewOutcome === 'COMPLETE' %}
+        <h3 class="govuk-heading-m">Confirming the questionnaire as complete makes the appeal available to inspectors </h3>
```
shepda, line 27: It looks like this text should use the `govukWarningText` component rather than an `h3`

packages/web-app/src/views/questionnaire-check-and-confirm.njk
```diff
@@ -0,0 +1,33 @@
+{% extends "layouts/main.njk" %}
```
shepda, line 1: This template should extent the form layout

packages/web-app/src/routes/index.js
```diff
@@ -27,6 +28,6 @@ router.use(`/${views.invalidAppealDetails}`, handleAppealAlreadyReviewed, invali
 router.use(`/${views.missingOrWrong}`, handleAppealAlreadyReviewed, missingOrWrongAppealDetails);
 router.use(`/${views.document}`, documentsServiceProxy);
 router.use(`/${views.appealAlreadyReviewed}`, appealAlreadyReviewed);
-router.use(`/${views.checkAndConfirm}`, checkAndConfirmDetails);
+router.use(`/${views.questionnairesForReview}/${views.checkAndConfirm}`, checkAndConfirm);
```
shepda, line 31: It looks like this is using the Appeals Check and Confirm router rather than the new Questionnaire one so as it is it'll be pointing to the wrong page?

We also want to keep the Appeals Check and Confirm page so the Questionnaire one should be added rather than replacing it.

demzzydev: yeah this was a mistake, will add back.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
shepda: The HTML can be used from the prototype but we should check it before using it in the app.

Looking at the prototype that text isn't there, although it is in the screenshot so probably need to check this with Anthony but assuming that we need it then it should use this component as it'll give the text the correct styling including the exclamation mark icon.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
shepda: It still needs to do a POST though because AC2 has it updating the status and updating the record in the database.

demzzydev: Fair enough

packages/web-app/src/views/questionnaire-check-and-confirm.njk
demzzydev: True, I'll contact him

packages/web-app/src/middleware/check-appeal-outcome.spec.js
```diff
@@ -0,0 +1,39 @@
...
+      title: 'Return next if there is appeal id in the params',
+      req: {
+        ...mockReq,
+        session: {},
+        params: {
+          appealId: 12345,
+        },
+      },
+      expected: (req, res, next) => {
+        expect(next).toBeCalledTimes(1);
```
shepda, line 27: We could include an assertion here to check that `req.session.outcome = 'INCOMPLETE'`

packages/web-app/src/controllers/questionnaire-check-and-confirm.js
```diff
@@ -0,0 +1,147 @@
...
+  const mappedAddress = Object.values(address).map((item) => `${item} <br> `);
+  return mappedAddress.toString().replace(/,/g, '');
+};
+
+const compileMissingDocuments = (items) => {
+  const mappedDocuments = items.map((item) => `${item} <br>`);
+  return mappedDocuments.toString().replace(/,/g, '');
+};
+
+const getConfirmationSections = (appealReference, outcome) => {
```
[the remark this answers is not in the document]
demzzydev: As of right now it isn't needed but the purpose of these functions is to guide development when we have the ability to interface with the db, basically trying to front load as much development as possible within this frontend ticket.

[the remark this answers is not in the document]
shepda: This `appealReference` param is populated from the `appealId` param passed in the URL and the appeal id isn't going to be displayed on this page. The appeal reference however is and this can be passed in with the appeal data, currently in `req.session`. This can also be used to provide the Local planning department value.

For now and until we connect the UI to the db, we have mock data in https://github.com/Planning-Inspectorate/back-office/blob/main/packages/web-app/src/lib/api-wrapper.js as this will probably be the function that fetches the data and the idea is that this can be used to provide data to the app and by using the mock data and keeping the pages consistent we should be able to build the app mostly as we would like then hopefully only have to make minimal changes once the actual data is available.

If necessary this can be refactored when we have the actual data available.

packages/web-app/src/middleware/check-appeal-outcome.spec.js
shepda: The reason for the assertion is that in https://github.com/Planning-Inspectorate/back-office/pull/30/files#diff-e01496a939d3d50a4f28156c3afc14aac16ccb48a9f1614e0cb9081d0951d3d0R16 we have `req.session.outcome = 'INCOMPLETE';` which is going to happen when this function is called assuming there is an appeal id so the assertion is to check that this value is set when `next()` is called.

packages/web-app/src/controllers/questionnaire-check-and-confirm.js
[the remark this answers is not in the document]
demzzydev: I think that's the essence of what I'm saying, we can change it but it will be refactored again once we have the data available so is it worth doing in this PR when the aim is to get the frontend ready. 

Also are we saying that the appeal data is going to be readily available in req.session? Because that wasn't apparent when working on the ticket which is why the appealReference param is there in the first place.

packages/web-app/src/controllers/questionnaire-check-and-confirm.js
```diff
@@ -0,0 +1,147 @@
...
+  const payload = {
+    rows: [],
+  };
+
+  switch (outcome) {
+    case reviewOutcome.INCOMPLETE: {
+      payload.rows.push(
```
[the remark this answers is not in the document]
shepda: The address pages of the UI and the data structure that I could find suggests that they have the same structure.

For now we can leave it as it is and refactor in the future if we need to.

packages/web-app/src/middleware/check-appeal-outcome.spec.js
demzzydev: I understand that, the reason why this was set to incomplete was just me hardcoding the outcome, I would imagine that value would come from the db when we refactor this, so it wasn't done with any specific logic in place.

PR #36, opened 2021-10-19
feat(as-2859-2875-3125): appeal Review Complete pages

DavidJWhiteDev:


packages/web-app/src/views/review-complete.njk
```diff
@@ -0,0 +1,25 @@
...
+      </div>
+
+      <p class="govuk-body">{{ checkAndConfirmConfig.reviewCompleteText }}</p>
+
+      <h2 class="govuk-heading-m">What next?</h2>
+
+      <p class="govuk-body"><a href="/appeals-list" class="govuk-link">Return to task list</a></p>
+    </div>
+  </div>
+{% endblock %}
```
shepda, line 25: This page isn't a form so I don't think that it needs to extend the form layout, probably extending the two thirds page layout would be better otherwise the page will inherit other form-specific tags that come with the form layout.

However as the `content` block is being used here both the form and the two thirds page layouts are being skipped as the `content` block is defined in the main layout which is at the top level and this is why the grid layout tags have needed to be added. If the correct block is used then these will be inherited from the two thirds page layout. 

For the form layout the block to use is `formContent` and for the two thirds page layout the block to use is `pageContent`.

shepda (commented):
Also please check the unit tests as there is a failing test

PR #39, opened 2021-10-20
Subtask/as 3597 questionnaire already reviewed

demzzydev:
## Ticket Number
https://pins-ds.atlassian.net/browse/AS-3597

## Description of change
This ticket is strictly frontend, decided to leave out middleware which determines whether the questionnaire has already been reviewed for the backend portion of this ticket.

## Checklist

- [ ] Requires infrastructure changes
- [ ] If adding or remove environment variables (e.g. in `docker-compose.yaml`) then I have updated the appropriate Helm chart
- [ ] I have updated the documentation accordingly
- [x] My commit history in this PR is linear
- [x] New features have tests
- [ ] Breaking change (team conversation required)

## Important

Please do not merge from `master` (please only [rebase](https://github.com/foundry4/appeal-planning-decision/wiki/An-intro-to-Git-Rebase)). This keeps the history linear and easier to debug.

### DIFF
--- a/packages/web-app/src/config/views.js
+++ b/packages/web-app/src/config/views.js
@@ -1,5 +1,6 @@
 const views = {
   appealAlreadyReviewed: 'appeal-already-reviewed',
+  questionnaireAlreadyReviewed: 'questionnaire-already-reviewed',
   appealsList: 'appeals-list',
   questionnairesList: 'questionnaires-list',
   document: 'document',
--- a/packages/web-app/src/config/views.js
+++ b/packages/web-app/src/config/views.js
@@ -1,7 +1,6 @@
 const views = {
   appealAlreadyReviewed: 'appeal-already-reviewed',
   questionnaireAlreadyReviewed: 'questionnaire-already-reviewed',
-  questionnaireForReview: 'questionnaires-for-review',
   appealsList: 'appeals-list',
   questionnairesList: 'questionnaires-list',
   document: 'document',
--- a/packages/web-app/src/config/views.js
+++ b/packages/web-app/src/config/views.js
@@ -1,6 +1,7 @@
 const views = {
   appealAlreadyReviewed: 'appeal-already-reviewed',
   questionnaireAlreadyReviewed: 'questionnaire-already-reviewed',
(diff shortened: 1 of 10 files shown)


packages/web-app/src/controllers/questionnaire-already-reviewed.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable no-constant-condition */
```
shepda, line 1: It doesn't look like this eslint override is required?

packages/web-app/src/routes/index.js
```diff
@@ -31,5 +33,9 @@ router.use(`/${views.missingOrWrong}`, handleAppealAlreadyReviewed, missingOrWro
 router.use(`/${views.document}`, documentsServiceProxy);
 router.use(`/${views.appealAlreadyReviewed}`, appealAlreadyReviewed);
 router.use(`/${views.checkAndConfirm}`, checkAndConfirmDetails);
+router.use(
+  `/planning-inspectorate/appeals/${views.questionnaireAlreadyReviewed}/already-reviewed`,
```
shepda, line 37: This URL will likely have to be changed in the future, I see it's in the ticket but it's not consistent with the rest of the app.

It's also not the same as what's in the ticket, with `/planning-inspectorate/appeals/questionnaire-already-reviewed/already-reviewed` rather than `/planning-inspectorate/appeals/questionnaires-for-review/already-reviewed`

demzzydev: good spot, shilpa has asked for us to keep the URL's as they are in the ticket so that when shes does e2e tests they won't fail because of invalid URL's

packages/web-app/src/config/views.js
```diff
@@ -1,5 +1,7 @@
 const views = {
   appealAlreadyReviewed: 'appeal-already-reviewed',
+  questionnaireAlreadyReviewed: 'questionnaire-already-reviewed',
+  questionnaireForReview: 'questionnaires-for-review',
```
nssdesikan, line 4: @demzzydev I believe this file is used to define the views name. _questionnaireForReview_ does not seem to be a view

PR #60, opened 2021-11-01
fix(as-3564): implement new migrations for sql changes

rcmsoftwaredevelopment:
as-3564

implement new migrations for sql changes


shepda (approved):
LGTM

PR #65, opened 2021-11-03
feat(start appeal email): email sent on valid review outcome confirmation

biroldif:


packages/web-app/src/lib/get-address.js
```diff
@@ -0,0 +1,14 @@
...
+    siteAddress?.town,
+    siteAddress?.county,
+    siteAddress?.postcode,
+  ].filter((n) => n);
+
+module.exports = {
+  getAddress,
+  getAddressSingleLine: (siteAddress) => getAddress(siteAddress).join(', '),
+  getAddressMultiLine: (siteAddress) => getAddress(siteAddress).join('\n'),
+};
```
shepda, line 14: It appears that these methods are only used in the unit tests? If so, I think a better location for the file would be `test/utils`.

biroldif: i removed getAddress. getAddressSingleLine and getAddressMultiLine are used in notify.js

shepda: :+1:

shepda (approved):
LGTM

PR #69, opened 2021-11-03
feat(web-app): add breadcrumbs to questionnaire check and confirm page

swdpcomputing:
This feature cannot be merged into main before @nssdesikan 's ticket AS-3619


packages/web-app/src/controllers/questionnaire-check-and-confirm.js
```diff
@@ -114,29 +114,28 @@ const getConfirmationSections = (outcome) => {
...
+const getBreadcrumbs = () => [
   {
     text: 'Questionnaires for review',
-    href: '/planning-inspectorate/appeals/questionnaires-for-review',
+    href: '/questionnaires-list',
   },
   {
     text: 'APP/Q9999/D/21/1224115',
-    href: `/planning-inspectorate/appeals/questionnaires-for-review/review/${appealId}`,
+    href: `/review-questionnaire`,
```
nssdesikan, line 124: Please update this URL

packages/web-app/src/controllers/questionnaire-check-and-confirm.spec.js
```diff
@@ -51,15 +51,15 @@ describe('controllers/questionnaire-check-and-confirm', () => {
       const breadcrumbs = [
         {
           text: 'Questionnaires for review',
-          href: '/planning-inspectorate/appeals/questionnaires-for-review',
+          href: '/questionnaires-list',
         },
         {
           text: 'APP/Q9999/D/21/1224115',
-          href: `/planning-inspectorate/appeals/questionnaires-for-review/review/${appealId}`,
+          href: `/review-questionnaire`,
```
nssdesikan, line 58: Please update this URL

packages/web-app/src/controllers/questionnaire-check-and-confirm.spec.js
```diff
@@ -146,15 +146,15 @@ describe('controllers/questionnaire-check-and-confirm', () => {
...
```
nssdesikan, line 153: Please update this URL

nssdesikan (changes requested):
_Main_ needs to be rebased onto this branch for

swdpcomputing (commented):
LGTM

packages/web-app/src/views/questionnaire-check-and-confirm.njk
```diff
@@ -41,7 +60,7 @@
     value: { text: questionnaireData.localPlanningAuthorityName }
 } %}
 
-{% if reviewOutcome === '2' %}
+{% if reviewOutcome === 'Incomplete' %}
```
demzzydev, line 63: Why has this been changed? Or we mapping this somewhere?

packages/web-app/src/views/questionnaire-check-and-confirm.njk
```diff
@@ -50,7 +69,7 @@
 {% block formContent %}
     <h2 class="govuk-heading-l">Check and confirm</h2>
     {{ govukSummaryList({ rows: sections }) }}
-    {% if reviewOutcome === '3' %}
+    {% if reviewOutcome === 'Complete' %}
```
demzzydev, line 72: We need to look at why this has been changed as well.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
nssdesikan: This is the value that is being passed from review-questionnaire-submission page. This is how- Incomplete / Complete - it was developed originally. But recent change - not by me - changed this to 1, 2 similar to they do in Team 1.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
nssdesikan: This is the value that is being passed from review-questionnaire-submission page. This is how- Incomplete / Complete - it was developed originally. But recent change - not by me - changed this to 1, 2 & 3 similar to they do in Team 1.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
demzzydev: I imagine the reason why it was changed to a numeric value is too map to the PK of the LookUpQuestionnaireOutcome table. Might be worth reaching out to Team 1 on confirming why do they do it in the first place so we can be consistent across the board otherwise we risk breaking things unnecessarily.

shepda: I agree with @demzzydev, this value shouldn't be changed.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
shepda: I agree with @demzzydev, this value shouldn't be changed.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
shepda: If the form is passing `Incomplete` as a value then that's the value that needs to be changed

packages/web-app/src/views/questionnaire-check-and-confirm.njk
shepda: If the form is passing `Complete` as a value then that's the value that needs to be changed

shepda (commented):
8 of these files have had the blank line at the end of the file removed and this will cause linting errors. Please check and add this back in again.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
nssdesikan: @shepda I can change it but what is the difference in sending the value as '1' or 'Complete'? Is it for db purpose?

nssdesikan:
> blank line at the end of the file removed

@shepda  Blank line removed at the end of those files in 'Main'. Should I still revert my changes in those files?

shepda:
> > blank line at the end of the file removed
> 
> @shepda Blank line removed at the end of those files in 'Main'. Should I still revert my changes in those files?

In most of the files the blank line at the end is the only change so those files will be removed from the PR. The other files that have other changes are fine to keep those changes.

packages/web-app/src/views/questionnaire-check-and-confirm.njk
nssdesikan: Reverted the changes as per PR comments

packages/web-app/src/views/questionnaire-check-and-confirm.njk
nssdesikan: Reverted the changes as per PR comments

shepda (approved):
LGTM

PR #84, opened 2021-11-10
Feature/as 3122 saving to db

shepda:
1. Update the api to save has appeal data with a stored procedure. This accepts a json string as a param, fetched the latest record, merges the json data with the latest record and saves as a new record. The trigger then makes the new record the latest record.

2. Update the appeals journey pages in the web-app to save data after each page submission. The json created after each page submission needs to have certain field names to match what the stored procedure is expecting and this mapping is handled by `src/config/db-fields.js`. Any additional fields in the json that the stored procedure isn't expecting will be ignored.


packages/api/database/migrations/00025-create-create-has-appeal-procedure.js
```diff
@@ -0,0 +1,193 @@
...
+        @inspectorFirstName = InspectorFirstName,
+        @inspectorSurname = InspectorSurname,
+        @inspectorId = InspectorID,
+        @inspectorSpecialismId = InspectorSpecialismID,
+        @scheduledSiteVisitDate = ScheduledSiteVisitDate,
+        @decisionOutcomeId = DecisionOutcomeID,
+        @decisionLetterId = DecisionLetterID,
+        @decisionDate = DecisionDate,
+        @descriptionDevelopment = DescriptionDevelopment
+      FROM HASAppeal
```
rcmsoftwaredevelopment, line 59: Can this be done with (nolock)

rcmsoftwaredevelopment (approved):

packages/api/database/migrations/00025-create-create-has-appeal-procedure.js
shepda: Sure, will do

rcmsoftwaredevelopment (approved):

PR #92, opened 2021-11-16
feat(as-3690): set outcome for the questionnaire review

demzzydev:
feat(as-3690): setting outcome with db integration


swdpcomputing (changes requested):
Please rebase commits down to one commit.

swdpcomputing (approved):

PR #99, opened 2021-11-19
fix(as-3039): populate user selected/entered data during page navigation

nssdesikan:
This fix populates the values that the CO selected / Entered in the Review Questionnaire page whenever they navigated back to this page


swdpcomputing (commented):
These changes don't seem to reflect the ticket requirements:

Add text length validation for the text boxes.

Add text for the listed building section to be passed to the next screen.

swdpcomputing (approved):

PR #104, opened 2021-11-22
feat(appeal details): page implementation completed

biroldif:


packages/web-app/src/views/appeal-details.njk
```diff
@@ -0,0 +1,396 @@
...
+{% from "govuk/components/accordion/macro.njk" import govukAccordion %}
+{% from "components/document-link.njk" import documentLink %}
+
+{% block formContent %}
+
+    <style>
+        .govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__key, .govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__value, .govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__actions {
+            border-bottom: 0;
+        }
+    </style>
```
shepda, line 17: This styling would be better in a separate scss file rather than inline

packages/web-app/src/views/appeal-details.njk
```diff
@@ -0,0 +1,396 @@
...
+
+    <h2 id="summary" class="govuk-heading-l section-heading">Summary</h2>
```
shepda, line 19: The prototype has a blue border above this heading, so we need that here?

packages/web-app/src/views/appeal-details.njk
```diff
@@ -0,0 +1,396 @@
...
+        }, true) }
+        },
+        {
+          key: { text: "Local planning department" },
+          value: { text: appealData.localPlanningAuthorityName }
+        }
+      ]
+    }) }}
+
+    <h2 id="case-file" class="govuk-heading-l govuk-!-margin-top-9 section-heading">Case File</h2>
```
shepda, line 47: The prototype has a blue border above this heading, so we need that here?

packages/web-app/src/views/appeal-details.njk
```diff
@@ -0,0 +1,396 @@
...
+                }
+              ]
+            })
+            }
+          }
+        ]
+      }) }}
+    </div>
+
+    <h2 id="stages" class="govuk-heading-l govuk-!-margin-top-9 section-heading">Stages</h2>
```
shepda, line 287: The prototype has a blue border above this heading, so we need that here?

packages/web-app/src/config/review-appeal-submission.js
```diff
@@ -6,6 +6,11 @@ const reviewOutcomeOption = {
   incomplete: '3',
 };
 
+const reviewOutComeById = Object.keys(reviewOutcomeOption).reduce((result, current) => {
+  const key = reviewOutcomeOption[current];
+  return Object.assign(result, { [key]: current });
+}, {});
```
shepda, line 12: Not entirely sure we need to map the ids to a string value to use as a conditional in the code, we could just refer to the id as we've done in other places, it might be a bit more readable right now but if the db value changes then the meaning of this string value could change or be lost unless we go through the UI to change them. More of an observation rather than expecting it to be changed.

packages/web-app/src/views/appeal-details.njk
```diff
@@ -0,0 +1,396 @@
+{% extends "layouts/form.njk" %}
```
shepda, line 1: Doesn't look like this template needs to be a form so could extend `two-thirds-page`?

packages/web-app/src/sass/main.scss
```diff
@@ -4,3 +4,12 @@
 a {
   @extend .govuk-link;
 }
+
+.govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__key, .govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__value, .govuk-accordion__section-content .govuk-summary-list__row:last-child .govuk-summary-list__actions {
+  border-bottom: 0;
+}
+
+h2.section-heading {
+  border-top: 3px solid #1d70b8;
```
shepda, line 13: Looks like the prototype has the border at 2px rather than 3px.

Also we could use the GOVUK variable `$govuk-brand-colour` for the blue colour rather than specifying it explicitly as this would keep consistency with the other colours.

shepda (approved):
LGTM

PR #105, opened 2021-11-22
fix(as-3680): search appeal

rcmsoftwaredevelopment:
fix(as-3680): search appeal

### DIFF
--- a/packages/api/src/controllers/appeal-search.js
+++ b/packages/api/src/controllers/appeal-search.js
@@ -1,0 +1,25 @@
+/* istanbul ignore file */
+const { searchAppeal } = require('../lib/db-wrapper');
+const logger = require('../lib/logger');
+const ApiError = require('../lib/api-error');
+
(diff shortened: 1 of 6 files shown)


packages/api/src/controllers/appeal-search.js
```diff
@@ -0,0 +1,25 @@
+/* istanbul ignore file */
+const { searchAppeal } = require('../lib/db-wrapper');
+const logger = require('../lib/logger');
+const ApiError = require('../lib/api-error');
+
+const appealSearch = async (req, res) => {
+  try {
+    const { searchString } = req.params;
+
+    console.log(searchString);
```
shepda, line 10: So we don't get this logging statement in the prod logs, I think it would be better to either remove the `console.log()` or change it to a debug statement, for example `logger.debug({ seatchString} 'Search string');` so we'd only get it in a dev env.

shepda (approved):
LGTM

PR #117, opened 2021-12-01
feat(update appeal decision date): page implementation

biroldif:


packages/web-app/src/controllers/appeal-details/application-decision-date.js
```diff
@@ -0,0 +1,83 @@
...
+      currentPage,
+      nextPage: `${appealDetails}/${appealId}`,
+      viewData: casework,
+      saveData: saveAppealSubmissionData,
+    });
+  } catch (e) {
+    logger.error(e);
+    const options = viewData(casework, errors, [{ text: e.toString(), href: '#' }]);
+    res.render(currentPage, options);
+  }
```
shepda, line 82: `saveAndContinue()` shouldn't need to be wrapped in a try/catch block as it already handles this.

packages/web-app/src/controllers/appeal-details/application-decision-date.js
```diff
@@ -0,0 +1,83 @@
...
+          month: body['decision-date-month'],
+          year: body['decision-date-year'],
+        },
+      },
+      errors,
+      errorSummary
+    );
+    res.render(currentPage, options);
+    return;
+  }
```
shepda, line 59: Errors shouldn't need to be handled here as `saveAndContinue()` already handles this

packages/web-app/src/controllers/appeal-details/application-decision-date.spec.js
```diff
@@ -0,0 +1,168 @@
...
+          },
+        },
+        errorSummary: [],
+        errors: {
+          key: 'value',
+        },
+      });
+    });
+
+    it('should handle exceptions correctly', async () => {
```
shepda, line 131: We shouldn't need a test to cover this scenario as `saveAndContinue()` handles this scenario and already has tests to cover it.

packages/web-app/src/middleware/combine-date-input.spec.js
```diff
@@ -0,0 +1,87 @@
...
+    it(description, () => {
+      const next = jest.fn();
+      const req = given();
+
+      combineDateInputs(req, mockRes(), next);
+
+      expected(req, mockRes(), next);
+    });
+  });
+});
```
shepda, line 87: Day and month values without a leading 0 are valid values for these fields so perhaps we should add a test for this scenario too?

packages/web-app/src/views/appeal-details/application-decision-date.njk
```diff
@@ -0,0 +1,54 @@
+{% extends "layouts/two-thirds-page.njk" %}
```
shepda, line 1: This template should extend the form layout as this will reduce the code required in this template

packages/web-app/src/views/appeal-details/application-decision-date.njk
```diff
@@ -0,0 +1,54 @@
...
+{% from "govuk/components/date-input/macro.njk" import govukDateInput %}
+
+{% block pageContent %}
+
+  {% if errors %}
+    {{ govukErrorSummary({
+      titleText: "There is a problem",
+      errorList: errorSummary
+    }) }}
+  {% endif %}
```
shepda, line 15: The `govukErrorSummary` should already be included as part of the two-thirds-page layout

packages/web-app/src/views/appeal-details/application-decision-date.njk
```diff
@@ -0,0 +1,54 @@
...
+          value: appealData.decisionDate.year
+        }
+      ]
+    } ) }}
+
+    {{ govukButton({
+      text: "Continue",
+      type: "submit",
+      attributes: { "data-cy":"button-save-and-continue"}
+    }) }}
```
shepda, line 51: We have a custom button component which wraps the `govukButton` so would probably be better to use here

packages/web-app/src/routes/appeal-details/application-decision-date.js
```diff
@@ -0,0 +1,22 @@
...
+const combineDateInputsMiddleware = require('../../middleware/combine-date-inputs');
+const expressValidationErrorsToGovUkErrorList = require('../../lib/express-validation-errors-to-govuk-error-list');
+
+const router = express.Router();
+
+router.get('/', applicationDecisionDateController.getApplicationDecisionDate);
+router.post(
+  '/',
+  [
+    combineDateInputsMiddleware,
```
shepda, line 15: Rather than calling `combineDateInputsMiddleware` as middleware so it needs to iterate all the form fields to check for a date field, couldn't it be a function which gets called when required for a specific date field, perhaps in the controller or wherever the date combining is needed?

shepda (approved):
LGTM

PR #134, opened 2021-12-07
Fix/as 4000 data not removed

shilpabasu22:


e2e-tests/cypress/support/db-queries/select-case-reference-from-db.js
```diff
@@ -1,6 +1,7 @@
 export const selectCaseReferenceFromDb = (status) =>{
     if(status=== 'Received'){
-        cy.sqlServer("select caseReference from [QuestionnaireData] where questionnaireStatus='RECEIVED'").then( (res)=>{
+        cy.sqlServer("select * from [QuestionnaireData] where questionnaireStatus='RECEIVED'").then( (res)=>{
+            console.log(res);
```
demzzydev, line 4: Do we need to log this?

demzzydev (approved):
LGTM

karunak21 (approved):
all good

PR #136, opened 2021-12-08
Subtask/as 4034 format data save questionnaire review

nssdesikan:


packages/web-app/src/controllers/review-questionnaire-submission.js
```diff
@@ -243,6 +264,35 @@ const getValues = (body) => {
   return values;
 };
 
+const getValidSelectedMissingOrIncorrect = (values) => {
```
swdpcomputing, line 267: This function is very complex and long. Can it be split up and documented js.doc please?

swdpcomputing (changes requested):

PR #145, opened 2022-03-07
Feature/bocm 18 basic state machine

mayya-bond-kc:


radum:
Should we keep the folder names in `state-machine` format rather than using _?

mayya-bond-kc:
@radum once we have the functionality more fleshed out I will be breaking things out into their own files and rename the folders to make sure things make sense

PR #146, opened 2022-03-08
Add api request libraries

radum:
I have added an example for the validation flow on how the routing, controller, and service for calling the API should work.

### DIFF
--- a/apps/web/.env.development
+++ b/apps/web/.env.development
@@ -24,5 +24,5 @@ PORT=3000
 # Default : 3443
 HTTPS_PORT=3443
 
-# API Endpoint
-API_ENDPOINT=http://localhost:3000
+# API specific vars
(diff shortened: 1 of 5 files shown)


apps/web/.env.development
```diff
@@ -23,3 +23,6 @@ HTTPS_ENABLED=false
 PORT=3000
 # Default : 3443
 HTTPS_PORT=3443
+
+# API Endpoint
+API_ENDPOINT=http://localhost:3000
```
mayya-bond-kc, line 28: `API_ENDPOINT` is not super clear, maybe rename to `API_HOST`?

mayya-bond-kc (approved):
LGTM

apps/web/.env.development
radum: I will update.

PR #148, opened 2022-03-08
[BOCM-20] adds Dockerfile for API

mayya-bond-kc:


apps/api/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:16-alpine AS base
+WORKDIR /src/app
+COPY package*.json ./
+
+FROM base AS dependencies
+
+COPY packages/planning-inspectorate-libs packages/planning-inspectorate-libs
+
+COPY ./apps/api .
```
radum, line 9: Shouldn't we keep the folder structure to make sure npm workspaces are working properly?

So that the CMD will be node `apps/api/...`

mayya-bond-kc: Can do!

radum (approved):
Small stuff.

`npm ci` to use the lock file, and I don't think --only exists anymore, but if we set NODE_ENV=production it will do what you intent it to do.

PR #151, opened 2022-03-09
BOCM-12 - Validation office landing page

hamidhazrati:


apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -1,13 +1,52 @@
...
 
+/**
+ * Given an item object transform it into - govuk/components/table/macro.njk" import govukTable format
+ *
+ * @param {*} item to do
+ * @returns {any} to do
+ */
+function makeGovTableRow(item){
+	return [
+		{ html: '<a href="#">'  + item.AppealReference + '</a>' },
```
radum, line 11: Please don't build html in the server code. It should be done in the NJK files.

apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -1,13 +1,52 @@
 import { findAllNewAppeals } from './validation.service.js';
 
+/**
+ * Given an item object transform it into - govuk/components/table/macro.njk" import govukTable format
+ *
+ * @param {*} item to do
+ * @returns {any} to do
+ */
+function makeGovTableRow(item){
```
radum, line 9: I think this function can be part of the actual controller function. Is not used anywhere else. I can even argue that this could be refactored to be the function of a reducer traversing the appeals data, but even so should be part of the controller.

apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -1,13 +1,52 @@
 import { findAllNewAppeals } from './validation.service.js';
 
+/**
+ * Given an item object transform it into - govuk/components/table/macro.njk" import govukTable format
+ *
+ * @param {*} item to do
```
radum, line 6: If you are not use JSDoc please don't add them. Is not a requirement for now.

apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -1,13 +1,52 @@
...
+ * @returns {void} to do
+ */
 async function getValidationDashboard(request, response) {
+
 	const appealsListData = await findAllNewAppeals();
 
+	const incompleteAppeals = [];
+	const newAppeals = [];
+
+	for (const item in appealsListData ) {
```
radum, line 31: I would have used reduce. But that is just me :).

apps/web/src/server/views/_includes/components/InfoBox/InfoBox.njk
```diff
@@ -0,0 +1,7 @@
+{% macro InfoBox(message) %}
+
+	<div class="pi-info-box govuk-table govuk-!-padding-9">
```
radum, line 3: Not sure why you used `govuk-table` but this is not a table and it just makes the code confusing, just manually add the code you need.

apps/web/src/server/views/validation/dashboard.njk
```diff
@@ -1,44 +1,45 @@
...
-                        No new appeals
-                    </td>
-                </tr>
-            </tbody>
-        {% endif %}
-    </table>
-{% endblock %}
+{% from "_includes/components/InfoBox/InfoBox.njk" import InfoBox %}
+
+<% set head = [
```
radum, line 4: Can we rename this to `tableHeadItems` or something similar to make it more clear what is its intent.

apps/web/src/server/views/validation/dashboard.njk
```diff
@@ -1,44 +1,45 @@
...
+			{ text: "Appeal reference" },
+			{ text: "Received on" },
+			{ text: "Appeal site" }
+			]
```
radum, line 8: Is the indentation off here?

apps/web/src/server/views/validation/dashboard.njk
```diff
@@ -1,44 +1,45 @@
...
+%>
+
+	{% block content %}
+
+		{% from "govuk/components/table/macro.njk" import govukTable %}
```
radum, line 13: Should be on top of the file like the other one.

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,14 @@
+@use "govuk/settings/_colours-applied" as colors;
+
+$pi-govuk-background-colour:  colors.govuk-colour("light-grey");
```
radum, line 3: You don't need a sass var for this, you can just call the Sass function where it is used.

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,14 @@
...
+$pi-color: rgb(11, 12, 12);
```
radum, line 4: Same, there is no need for a Sass var as we are not reusing this anywhere.

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,14 @@
...
+
+.pi-info-box {
+	background-color: $pi-govuk-background-colour;
+	border-top: solid colors.$govuk-border-colour 1px;
+	border-bottom: solid colors.$govuk-border-colour 1px;
+	color: $pi-color;
```
radum, line 10: We should use the values from the design system. If its not there, use the closest match.

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,14 @@
...
+	display: block;
+	padding: 10px;
```
radum, line 12: We have utility classes for this in the [DS](https://design-system.service.gov.uk/styles/spacing/). So remove it from here and use a class.

radum: Although I think you already did so not sure what this is for.

radum (changes requested):

radum:
@hamidhazrati You also have conflicts so please rebase your branch before tackling any of the PR review comments.

apps/web/src/server/app/validation/validation.controller.js
hamidhazrati: I am surprised at this as I thought this is exactly what we had agreed to do. It surprised me at the time. But will move this out.

apps/web/src/server/app/validation/validation.controller.js
hamidhazrati: Yes it originally was a foreach reducer , but lint complained and QuickFix on IDE made it into a function. Will move back and put a disable lint comment against it

apps/web/src/server/app/validation/validation.controller.js
hamidhazrati: 👍

apps/web/src/server/app/validation/validation.controller.js
hamidhazrati: see previous comments about lint errore
👍

apps/web/src/server/views/_includes/components/InfoBox/InfoBox.njk
hamidhazrati: 👍

apps/web/src/server/views/validation/dashboard.njk
hamidhazrati: 👍

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
hamidhazrati: 👍

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
hamidhazrati: 👍

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
hamidhazrati: you need to talk me through this ...

apps/web/src/server/app/validation/validation.controller.js
radum: Not for each, I meant `array.reduce()` but its OK it was just a thought about it, you can keep the for loop.

apps/web/src/server/app/validation/validation.controller.js
radum: Perhaps something was lost in the comms and I apologies if that is the case but I would have never said to put html in the server code.

apps/web/src/server/app/validation/validation.controller.js
radum: Thats ok, you can keep it as is.

apps/web/src/styles/7-components/InfoBox/InfoBox.scss
radum: In the NJK file of the component you already have `<div class="pi-info-box govuk-table govuk-!-padding-9">` padding added via `govuk-!-padding-9`.

Why do you need to add it again here?

apps/web/src/server/app/validation/validation.controller.js
hamidhazrati: moved into main function

PR #152, opened 2022-03-10
Feature/bocm 37 endpoint for validation appeal info

MariaValeroGonzalezKC:


apps/api/src/server/app.js
```diff
@@ -22,6 +23,8 @@ app.use('/', homeRoutes);
 
 app.use('/validation', validationRoutes);
 
+app.use('/appeal-review', appealReviewRoutes);
+
```
mayya-bond-kc, line 27: I feel like we can use the validation routes here. So instead of adding a new folder, that we add a new function to the validation controller to return the appeal details that we need to validate

apps/api/src/server/app/validation/validation.controller.js
```diff
@@ -15,6 +15,25 @@ const appealsList = [
...
+		AppealId : 1,
+		AppealReference: 'APP/Q9999/D/21/1345264',
+		AppellantName: 'Lee Thornton',
+		AppealStatus:'new',
+		Received: '23 Feb 2022',
+		AppealSite:'96 The Avenue, Maidstone, Kent, MD21 5XY',
+		LocalPlanningDepartment: 'Maindstone Borough Council',
+		PlanningApplicationReference: '48269/APP/2021/1482'
+	}
+];
```
mayya-bond-kc, line 29: Just one last point, I don't believe we need to return an array here. I think we will only ever need to return a hash, so we can remove those square braces.

apps/api/src/server/app/validation/validation.test.js
```diff
@@ -19,3 +19,19 @@ test('gets the appeal information', async (t) => {
...
+		AppealReference: 'APP/Q9999/D/21/1345264',
+		AppellantName: 'Lee Thornton',
+		AppealStatus:'new',
+		Received: '23 Feb 2022',
+		AppealSite:'96 The Avenue, Maidstone, Kent, MD21 5XY',
+		LocalPlanningDepartment: 'Maindstone Borough Council',
+		PlanningApplicationReference: '48269/APP/2021/1482'
+	};
+	t.is(resp.status, 200);
+	t.deepEqual(resp.body, [appealReviewInfo]);
```
mayya-bond-kc, line 36: So once we remove the square braces in `validation.controller.js`, then we will need to change this line to `t.deepEqual(resp.body, appealReviewInfo)`

mayya-bond-kc (approved):

PR #154, opened 2022-03-11
Feature/bocm 38 add sql server integration

mayya-bond-kc:


apps/api/package.json
```diff
@@ -30,7 +33,10 @@
     "log-process-errors": "8.0.0",
     "morgan": "1.10.0",
     "planning-inspectorate-libs": "*",
-    "xstate": "4.30.3"
+    "xstate": "4.30.3",
+    "sequelize": "6.17.0",
+    "tedious": "14.3.0",
+    "sequelize-cli": "6.4.1"
```
radum, line 39: If you are using npx above in the script, no need to have it as a dependency.

apps/api/src/config/database-config.js
```diff
@@ -0,0 +1,43 @@
+import { loadEnvironment } from 'planning-inspectorate-libs';
+
+loadEnvironment(process.env.NODE_ENV);
+
+export default {
+	development: {
```
radum, line 6: The values cam change based on the NODE_ENV var and `loadEnvironment` does that for you. I don't think you need to replicate them like this for each env.

apps/api/src/db/migrations/20220310140939-create-appeal.cjs
```diff
@@ -0,0 +1,26 @@
...
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE
+      }
+    });
+  },
+  async down(queryInterface, Sequelize) {
+    await queryInterface.dropTable('Appeals');
+  }
+};
```
radum, line 26: You are missing a new line at the end of the file here.

apps/api/src/server/app.js
```diff
@@ -4,14 +4,12 @@ import morgan from 'morgan';
 import helmet from 'helmet';
 import { homeRoutes } from './app/home/home.routes.js';
 import { validationRoutes} from './app/validation/validation.routes.js';
-
-
-const app = express();
-
 import swaggerUi from 'swagger-ui-express';
```
radum, line 7: The order of imports should be on top all that come from node_modules, then our own files.

apps/api/.env.development
```diff
@@ -20,3 +20,10 @@ BUNDLE_ANALYZER=false
 PORT=3000
 # Default : 3443
 HTTPS_PORT=3443
+
+# - - - - - - - - - Database - - - - - - -
+SQL_SERVER_HOST=0.0.0.0
```
radum, line 25: Does have to be the 0.0.0.0 and listen on all network interfaces? Or is the loopback 127.0.0.1 enough for local dev? Out of curiosity.

mayya-bond-kc: I haven't tested out with 127.0.0.1. both should work fine

radum: The thing is that on 0.0.0.0 all your network will be able to see your server running. Not that might be such a big issue. But was curious if it was a deliberate choice.

apps/api/src/server/app/repositories/appeal.repository.test.js
```diff
@@ -0,0 +1,18 @@
+import test from 'ava';
```
AnthonyJWalsh, line 1: Why are we using and what is "ava" ? I thought the plan was to use jest for testing?

radum: I suggested to Mayya as an alternative as Jest is not really maintained anymore and it also doesn't support ESM. Jest is mostly used for Web stuff and not necessarily for Node.

[Ava](https://github.com/avajs/ava) is an alternative that is designed for Node, runs tests concurrently and is as easy to use as any other testing library I guess.

If the preferred lib is Jest we can convert back to it as the refactor is minimal.

PR #155, opened 2022-03-11
/feature/Bocm 39 creation of endpoint patch

MariaValeroGonzalezKC:


apps/api/src/server/app/validation/validation.controller.js
```diff
@@ -36,4 +37,14 @@ const getValidation = function (request, response) {
 	response.send(appealsList);
 };
 
-export { getValidation, getAppealReview };
+const updateValidation = function (request, response) {
+	console.log(request.text);
+	const errors = validationResult(request);
+	if (!errors.isEmpty()) {
+		console.log(errors);
```
mayya-bond-kc, line 44: In the future, I think we can remove these `console.log`

mayya-bond-kc (approved):
Looks good to me!

apps/api/src/server/app/validation/validation.controller.js
MariaValeroGonzalezKC: will do in the next PR

PR #195, opened 2022-03-24
BOCM-52 - Case officer review page

james-hill-kc:
## Describe your changes

Template and controller code for LPA review page

[BOCM-52](https://pins-ds.atlassian.net/browse/BOCM-52)

## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/app/lpa/lpa.controller.js
```diff
@@ -26,3 +25,154 @@ export async function getLpaDashboard(request, response, next) {
...
+/**
+ * GET the questionnaire review page
+ * It will fetch the questionnaire details for the specified ID and will render the review form.
+ *
+ * @param {import('express').Request} request - Express request object
+ * @param {import('express').Response} response - Express request object
+ * @param {Function} next  - Express function that calls then next middleware in the stack
+ * @returns {void}
+ */
+export async function getReviewQuestionnaire(request, response, next) {
```
brpins, line 38: Just leave this as is, as I'll likely update all request handler signatures to a stronger-typed custom declaration at some point, but for info, a simpler way to js-document this function is:

/** @type {import('express').RequestHandler} */

(It also accepts generics)

apps/web/src/server/app/lpa/lpa.pipes.js
```diff
@@ -0,0 +1,69 @@
+import { body } from 'express-validator';
+
```
brpins, line 2: Shouldn't the validators in this file also be calling .trim() and .escape() to sanitize their input?

apps/web/src/server/app/lpa/lpa.pipes.js
```diff
@@ -0,0 +1,69 @@
...
+// All validation pipes will save into the current request all the validation errors that would be used
+// by the `expressValidationErrorsInterceptor` to populate the body with.
+
+/**
+ * Validate the questionnaire form to ensure any items marked as missing or incorrect have details provided, if there is a details field provided
+ *
+ * @returns {void}
+ */
+export const validateQuestionnairePipe = () => [
```
brpins, line 11: I realise this has been done elsewhere already, but I am not really sure why we are making these pipes into functions, only to call them without arguments. Leave as is I guess, just voicing curiosity!

apps/web/src/server/app/lpa/lpa.service.js
```diff
@@ -46,3 +46,11 @@ export async function findAllIncomingIncompleteQuestionnaires() {
 
 	return questionnairesListData;
 }
+
+export async function findQuestionnaireById(id) {
+	const data = await request(`case-officer/${id}`);
+
+	data.AppealSiteHtml = data.AppealSite ? data.AppealSite.replaceAll(',', '<br />') : '';
```
brpins, line 53: HTML manipulation shouldn't be in the service. You can just do this in nunjucks anyway, for example: {{ data.AppealSite | replace(",", "<br />")

apps/web/src/server/views/lpa/check-and-confirm.njk
```diff
@@ -0,0 +1,48 @@
...
+		serviceUrl: "/lpa",
+		containerClasses: "govuk-width-container"
+	}) }}
+{% endblock %}
+
+{% block content %}
+	{{ super() }}
+
+	<h1 class="govuk-heading-xl">
+		Review questionnaire
```
brpins, line 20: You also need to set the `{% block pageTitle %}` else this heading will not appear in the browser tab

apps/web/src/server/views/lpa/check-and-confirm.njk
```diff
@@ -0,0 +1,48 @@
...
+
+	{%- if reviewOutcome === 'complete' -%}
+		{{ govukWarningText({
+			iconFallbackText: "Warning",
+			text: "Confirming this questionnaire as complete makes the appeal visible to inspectors."
+		}) }}
+	{%- endif -%}
+
+	<div class="govuk-button-group">
+		{{ govukButton({ text: 'Confirm outcome' | default('Confirm outcome') }) }}
```
brpins, line 44: why do we need the` | default` filter? Isn't that just for when the initial text is undefined and it falls back to the default value?

apps/web/src/server/views/lpa/review-questionnaire.njk
```diff
@@ -0,0 +1,288 @@
...
+						fields.plansUsedToReachDecision.error,
+						{ html: detailsInput('textarea', 'plans-used-to-reach-decision', 'Which plans are missing?', fields.plansUsedToReachDecision.details)}
+					)}
+				]
+			]
+		}) }}
+
+		<h2 class="govuk-heading-l govuk-!-margin-top-9">Local plans and policies</h2>
+
+		{{ govukTable({
```
brpins, line 145: This table is full of hardcoded documents right now... is that intentional?

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: Possibly, although I'd say it would be better to sanitize the values themselves before they are checked by the validator. This would be a good point to discuss with Radu when he's back as it's his setup, I was just trying to keep things consistent with the pre-existing code in the validation journey pipes file.

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: It's a good point! We can discuss with Radu when he's back

apps/web/src/server/app/lpa/lpa.pipes.js
brpins: Yeah, this is what I mean. The inputs should be sanitized/trimmed first, else notEmpty could pass for a string with an empty space, I guess. e.g.

```
body('someprop')
  .escape()
  .trim()
  .notEmpty()
  ...
```
I see there's not really any escaping/trimming being done, so you can leave it as we'll have to address the rest of the validators anyway.

apps/web/src/server/app/lpa/lpa.service.js
james-hill-kc: A good point! The data format for the AppealSite has changed since I wrote this originally, so it would have needed updating anyway. I've refactored to move this out of the service into the controller.

apps/web/src/server/views/lpa/check-and-confirm.njk
james-hill-kc: I've added this to the 3 existing LPA templates

apps/web/src/server/views/lpa/check-and-confirm.njk
james-hill-kc: This was a result of copying templates where the text is passed in as a variable, I've removed the defaults from the LPA templates now

apps/web/src/server/views/lpa/review-questionnaire.njk
james-hill-kc: It was, the validation officer journey review-appeal template is the same currently. It seems as good a time as any to start adding this though, so I've added some code which renders these based on the documents data returned from the API (the data is still hardcoded in the API for now)

apps/web/src/server/lib/helpers.js
```diff
@@ -46,3 +47,24 @@ export const appealSiteObjectToText = (appealSite, separator) => {
...
 	return addressString;
 };
+
+/**
+ * Create an object of html strings containing file download links for consumption in nunjucks template, using the supplied documents data array from the LPA service
+ *
+ * @param {Array<object>} documents - array of document data items from API endpoint
+ * @returns {Object} - object containing html strings for consumption in nunjucks template
+ */
+export const makeDownloadLinksDataForTemplate = (documents) => {
```
james-hill-kc, line 57: Worth noting, I was debating putting this logic in the templates. I'm open to suggestions if anyone thinks it would be worth moving it there (probably as a custom filter)

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: Tagging @radum for visibility. Maybe we should add a ticket to cover updating the validation pipes for all journeys at once?

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: @radum another one for you to be aware of

apps/web/src/server/app/lpa/lpa.controller.js
```diff
@@ -1,6 +1,33 @@
...
+ * Create an array of row data for consumption in nunjucks template govukTable component, using the supplied data array from the LPA service
+ *
+ * @param {Array<object>} questionnairesList - array of questionnaire data items returned from findAllIncomingIncompleteQuestionnaires service method
+ * @returns {Array<object>} - array of row data objects for consumption in nunjucks template
+ */
+function makeQuestionnairesListRowsForTemplate (questionnairesList) {
+	const rows = [];
+
+	// eslint-disable-next-line unicorn/no-array-for-each
+	questionnairesList.forEach((item) => {
```
brpins, line 17: My personal opinion is formatting and html manipulation shouldn't be in the controller, as it becomes more like a ViewModel and and I don't think that's the design we're going for. I think controllers should just be handling the incoming request and rendering a template with data, then nunjucks does any template manipulation.

I would personally be passing the questionnaires list to the template, and then using a nunjucks macro or filter. Perhaps @radum has an opinion.

radum: I agree, although I didn't had strong feelings when we started due to the time we have, I think its a good thing to do from the start. Using macros should handle these situations, and if the data is way to skewed we can talk to the API devs.

apps/web/src/server/app/lpa/lpa.pipes.js
radum: I think I just copy pasted the example from the docs but true there is no reason for these to be functions, so you can change them here.

apps/web/src/server/lib/helpers.js
radum: I would say it is better as a template specific thing.

apps/web/src/server/views/lpa/review-questionnaire.njk
radum: The validation journey documents on the appeal details page are not hardcoded anymore.

apps/web/src/server/app/lpa/lpa.controller.js
james-hill-kc: Thanks guys, good to have clarity on the approach. I've refactored this to use a macro and it does seem cleaner. Is there a rule of thumb for when to manipulate data in the service/controller vs. in the template? For example, should any and all manipulation of data occur only in the templates, or just when it involves rendering strings/HTML?

I assume we will use this approach across the whole codebase. If so there are a few bits we should go back and refactor; I'll mark these with TODO comments.

james-hill-kc: After chatting with @radum I've refactored code in this PR which was generating html or otherwise performing template-specific formatting of the data. This involved adding a couple of macros and nunjucks helpers, but the approach does feel cleaner overall. I marked some other bits that were out of scope for this PR with TODO comments.

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: Updated all relevant pipes in the validation and case officer journeys with this. Let me know if you spot any I've missed.

apps/web/src/server/app/lpa/lpa.pipes.js
james-hill-kc: Updated!

apps/web/src/server/lib/helpers.js
james-hill-kc: This has been removed now in favour of makeDocumentLinksForType macro

mayya-bond-kc (approved):

apps/web/src/server/views/validation/review-appeal.njk
```diff
@@ -7,6 +7,16 @@
 {% from "app/components/document-list.component.njk" import documentList %}
 
 {% set isAppealIncomplete = appeal.AppealStatus === 'incomplete' %}
+{% from "_includes/macros/make-appeal-site-html.njk" import makeAppealSiteHtml %}
```
radum, line 10: Where is this used?

apps/web/src/server/views/validation/review-appeal.njk
```diff
@@ -7,6 +7,16 @@
...
+
+{# Macro use to generate a list of documents #}
+{% macro supportingDocumentsList(supportingDocuments)  %}
```
brpins, line 12: You can use the existing  documentList(supportingDocuments) for this

apps/web/src/server/views/lpa/review-questionnaire.njk
```diff
@@ -0,0 +1,268 @@
...
+			label: { text: labelText },
+			errorMessage: details.error and { text: details.error.msg },
+			spellcheck: true,
+			value: details.value,
+			rows: '3'
+		}) }}
+	{%- elif type === 'checkboxes' -%}
+		{{ govukCheckboxes({
+			idPrefix: idPrefix + '-missing-or-incorrect-reason',
+			name: idPrefix + '-missing-or-incorrect-reason',
```
radum, line 28: Any reason why we have a `-` in front of the name and id?

james-hill-kc: It's because the isPrefix is concatenated with the "missing or incorrect reason" text, without the `-` you would end up with IDs like `supporting-documentsmissing-or-incorrect-reason`

apps/web/src/server/views/validation/review-appeal.njk
james-hill-kc: I forgot to remove this after merging. The documentList macro didn't exist when I originally wrote this.

apps/web/src/server/views/_includes/macros/make-appeal-site-html.njk
```diff
@@ -0,0 +1,12 @@
+{% macro makeAppealSiteHtml(appealSiteData) %}
```
radum, line 1: Should we wrap the entire thing in an <address> tag to make it more semantic?

apps/web/src/server/app/lpa/lpa.controller.js
```diff
@@ -12,17 +14,166 @@ import { lpaRoutesConfig as routes } from '../../config/routes.js';
...
+				request.session.reviewWork.fields[camelCase(key.replace('-missing-or-incorrect-reason', ''))].details.error = { msg: errors[key].msg };
+			}
+		}
+
+		return response.render(routes.reviewQuestionnaire.view, {
+			backURL: `${routes.home.path}?direction=back`,
+			errors,
+			errorSummary,
+			questionnaireData,
+			// eslint-disable-next-line unicorn/consistent-destructuring
```
radum, line 132: Maybe this rule is too much. Please change it to a warning maybe in our eslint config and remove all this disable lines.

apps/web/src/server/views/_includes/macros/make-appeal-site-html.njk
james-hill-kc: I've removed this macro entirely as it's obsolete now (using `address` instead).

apps/web/src/server/app/lpa/lpa.controller.js
james-hill-kc: Very happily done! 

Personally I think most of these `unicorn` rules are unnecessary. They are very opinionated (for example https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md) and just serve to slow down development. Often the reasoning given for the rule is brief and vague and I don't see what makes [sindresorhus](https://github.com/sindresorhus) an authority on what parts of the javascript language should be summarily consigned to the dustbin.

I will admit the "prevent abbreviations" rule is a nice reminder. Just to show I'm not completely against it 😂 

Apologies for the complaining tone, I just feel quite strongly on the subject of linters and what I consider to be their rampant overuse!

radum (approved):

PR #213, opened 2022-03-30
Fix for incorrect data format being sent to API when missing documents are selected

james-hill-kc:
refactoring of validation controller postCheckAndConfirm to ensure correct format of data posted to backend

## Describe your changes

Moved code used to build Reason data object outside postCheckAndConfirm into its own method, refactored for clarity and extensibility.

NOTE: this also includes a find-and-replace fix for the typo "inflamatory" - @mayya-bond-kc tagging you for visibility.

?

## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -354,6 +354,59 @@ export function getCheckAndConfirm(request, response) {
...
+ * @returns {object} - resulting Reasons object in valid format (see /api-docs/#/default/post_validation__id_ for details)
+ */
+function formatOutcomeReasonsDataForApiPost(outcomeType, appealDetails) {
+	let data = {};
+
+	switch (outcomeType) {
+		case 'invalid':
+			data = {
+				// eslint-disable-next-line unicorn/no-array-reduce, unicorn/prefer-object-from-entries
+				...appealDetails.invalidReasons.reduce((previous, current) => {
```
brpins, line 371: Minor thing as there's no technical impact, but this naming is a bit confusing as it doesn't really match the signature of reduce. By its name, "current" would suggest a new representation of "previous", whereas it is in fact a key for the reason, while "previous" is the map being reduced.

james-hill-kc: I agree these could have been clearer (this is the terminology used in the reducer examples on MDN which is why I renamed the original single-letter variables to these). I've updated to "accumulator" and "value" now.

brpins (dismissed):

mayya-bond-kc (approved):

PR #215, opened 2022-03-31
BOCM-219 fix

james-hill-kc:
Fix for incorrect back url on validation check and confirm page (for invalid outcome)

## Describe your changes

Updated back URL for validation journey check and confirm page, when user has selected invalid outcome for the review

[BOCM-219](https://pins-ds.atlassian.net/browse/BOCM-219)

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/app/validation/validation.controller.js
+++ b/apps/web/src/server/app/validation/validation.controller.js
@@ -328,12 +328,14 @@ export function getCheckAndConfirm(request, response) {
 	let backURL;
 
 	if ('invalidAppealDetails' in appealWork) {
-		backURL = `/validation/${routes.reviewAppealRoute.path}/${appealData.AppealId}?direction=back`;
+		backURL = `/validation/${routes.invalidAppealOutcome.path}?direction=back`;
 	} else if ('incompleteAppealDetails' in appealWork) {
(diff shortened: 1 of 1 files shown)


apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -328,7 +328,7 @@ export function getCheckAndConfirm(request, response) {
 	let backURL;
 
 	if ('invalidAppealDetails' in appealWork) {
-		backURL = `/validation/${routes.invalidAppealOutcome.path}?direction=back`;
+		backURL = `/validation/${routes.reviewAppealRoute.path}/${appealData.AppealId}?direction=back`;
```
brpins, line 331: I don't think this change is correct. The back link _should_ take you to the invalid appeal outcome page.

The issue here is that the `changeOutcomeURL` is using whatever the back link is using, which is not correct, as the two destinations can differ.

james-hill-kc: Yes you're absolutely right. On closer review of the ACs these need to go to different destinations, which makes sense given they are separate links! This is fixed now.

brpins (approved):

PR #222, opened 2022-04-04
BOCM-35: Add incomplete appeal page

brpins:
BOCM-35: Add incomplete appeal page in validation
BOCM-46: Add changing of outcome for incomplete appeals
BOCM-196: Refactor nunjucks layouts and add three-column layout
BOCM-196: Add inspector domain / appeals details page

https://pins-ds.atlassian.net/browse/BOCM-35
https://pins-ds.atlassian.net/browse/BOCM-46
https://pins-ds.atlassian.net/browse/BOCM-196

## Type of change 🧩

- [ ] Documentation update


apps/web/src/styles/5-elements/_elements.page.scss
```diff
@@ -20,3 +20,7 @@ body {
 		scroll-behavior: smooth;
 	}
 }
+
+dd {
```
radum, line 24: I think this goes better in the generic folder in a file `_generic.reset.scss` perhaps.

apps/web/src/styles/main.scss
```diff
@@ -33,3 +34,9 @@
 
 // Shame file used for everytihng you will be shamed for.
 @use "shame";
+
+// FIXME: govuk accessible autocomplete – copied from
+// node_modules/accessible-autocomplete for whatever reason none of
+// "~accessible-autocomplete", "~accessible-autocomplete/src/autocomplete" or
+// "~accessible-autocomplete/src/autocomplete.css" will work
+@use "autocomplete.css";
```
radum, line 42: In `apps/web/scripts/compile-css.js` line 39, we should update that with the entire node_modules folder not just govuk-frontend and then we can import it directly. I will raise a PR and you can inherit the changes.

apps/web/src/styles/_shame.scss
```diff
@@ -0,0 +1,5 @@
+.pins-content--sticky {
+  position: -webkit-sticky;
```
radum, line 2: We use autoprefixer, so no need for this.

apps/web/src/styles/7-components/_table.scss
```diff
@@ -0,0 +1,10 @@
+.govuk-accordion__section-content {
+  .govuk-summary-list:last-child {
```
radum, line 2: We use tabs for indentation. Would be great if you update all required files with it.

Not sure why editorconfig didn't picked it up for you.

apps/web/src/styles/7-components/_related.scss
```diff
@@ -0,0 +1,13 @@
+
+@use "govuk/settings/_colours-applied" as colors;
+
+.pins-related {
+  border: 1px solid colors.$govuk-border-colour;
+  border-top: 4px solid colors.$govuk-brand-colour;
+  
+  // FIXME: No idea how to import these mixins without errors
```
radum, line 8: I will "fix" the imports from node_modules and I will provide an example.

apps/web/src/server/views/app/layouts/README.md
```diff
@@ -0,0 +1,16 @@
+# NJ Layouts
```
radum, line 1: The reason I used _ in front of the folders for layouts is to make the distinction between what can be used in express directly vs what can be extended or included only.

I think it provides better visual separation, but I have no strong feeling if you don't want to keep the _.

apps/web/src/server/lib/async-error-handler.js
```diff
@@ -0,0 +1,15 @@
+// @ts-check
+
+/**
+ * Wrap an asynchronous express middleware such that it catches and handles errors
+ *
```
radum, line 5: Do me a favour and leave a TODO here and say this should be evaluated again once we migrate to Express 5.

apps/web/src/server/app/app.express.js
```diff
@@ -92,8 +91,12 @@ const njEnvironment = nunjucks.configure(viewPaths, {
 	express: app // the express app that nunjucks should install to
 });
 
-njEnvironment.addFilter('stripQueryParamsDev', stripQueryParametersDevelopment);
-njEnvironment.addFilter('className', className);
+for (const filterName in nunjucksFilters) {
+	if (Object.prototype.hasOwnProperty.call(nunjucksFilters, filterName)) {
```
radum, line 95: You are checking here to see if we added this already?

apps/web/src/server/app/app.express.js
```diff
@@ -92,8 +91,12 @@ const njEnvironment = nunjucks.configure(viewPaths, {
 	express: app // the express app that nunjucks should install to
 });
 
-njEnvironment.addFilter('stripQueryParamsDev', stripQueryParametersDevelopment);
-njEnvironment.addFilter('className', className);
+for (const filterName in nunjucksFilters) {
```
radum, line 94: :) Smart loop. I love it.

apps/web/src/server/app/validation/__dev__/local-planning-departments.js
```diff
@@ -0,0 +1,2361 @@
+
+export default [
```
radum, line 2: This folder and file will be short lived until the API works right?

apps/web/src/server/app/inspector/inspector.router.js
```diff
@@ -0,0 +1,8 @@
+import express from 'express';
```
radum, line 1: The filename should be `inspector.route.js`.

apps/web/src/server/app/inspector/inspector.controller.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * View the appeal details as an inspector.
+ *
+ * @type {import('@pins/express').QueryHandler<AppealParams, ViewAppealDetailsRenderOptions>}
+ */
+export async function viewAppealDetails({ params }, response) {
+	const appeal = await findAppealById(params.appealId);
+
+	response.render('inspector/appeal-details', { appeal });
```
radum, line 23: I think we should add the route path and the views in the config file like we did in the validation, just in case we need to use them in multiple parts of our app.

apps/web/src/client/app.js
```diff
@@ -2,6 +2,7 @@
 //* This will be updated based on a router -> loader flow.
 
 import { initAll } from 'govuk-frontend';
+import './components/accessible-autocomplete';
```
radum, line 5: Move this to `apps/web/src/client/pages/default.js` please. And I will explain the reason behind it in a dev call.

apps/web/src/client/components/accessible-autocomplete.js
```diff
@@ -0,0 +1,12 @@
+import 'accessible-autocomplete';
+
+// on page load, progressively enhance all <select> elements with the
+// "accessible-autocomplete" class
+window.addEventListener('DOMContentLoaded', function () {
```
radum, line 5: We defer the script so there is no need for this. Also I think I will preload the JS and I will move it at the bottom.

apps/web/src/client/components/accessible-autocomplete.js
```diff
@@ -0,0 +1,12 @@
+import 'accessible-autocomplete';
```
radum, line 1: I would change this to a dynamic import and only loaded if the elements you are looking bellow exists on the page.

Something like this

```js
const accessibleAutocompleteEls = document.querySelectorAll('select.accessible-autocomplete');

if (accessibleAutocompleteEls.length > 0) {
  import('accessible-autocomplete').then(() => {
    [...accessibleAutocompleteEls].forEach((select) => {})
  });
}
```

apps/web/package.json
```diff
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "@hapi/boom": "9.1.4",
+    "accessible-autocomplete": "^2.0.4",
```
radum, line 28: We should lock the version by removing the ^ from all new deps.

apps/web/src/client/components/accessible-autocomplete.js
brpins: I see what you mean, but isn't the app.js just a single bundle of all the js in the app? A conditional import would only be worthwhile if this import can be fetched lazily? Or is that what happens?

apps/web/src/server/app/inspector/inspector.controller.js
brpins: So this was an intentional departure from the routes config way of doing things, both for path and view:

Regarding views, what is actually gained by maintaining an external map of paths to views? I'd argue it makes more sense that the controller function has ownership of  is the only place its rendered view should be declared. Furthermore, template names (and indeed paths) are not easily read without mentally constructing the view path by swapping between files.

Regarding path, importing the route path only really works if it doesn't need to be interpolated in any way. For example, imagine if you have in your config:

`path: 'appeals/:appealId/some-action'`

It is not really possible to use the config approach now, because the different parts of the url require a dynamic id to be inserted in the middle, so some kind of string interpolator would be needed. Alternatively, you could maintain one path as two different entries in the config and start combining them to form urls, but that too doesn't make sense: I saw we were already doing things like `${config.routes.somePath}/${appealId}` in the controllers. That already seems like an anti-pattern because it assumes you already know what the construction of the path without referring to the config at all. Likewise, we are manually prepending `/validation` as a string to the routes in the controller, which is also contradictory to the idea of the config pattern.

Overall, I feel like like the config approach has too much overhead for no real benefit. My preference would be that the <domain>.route.js is the source of truth of routes, and their usage is akin to typing out the name of a constant. If you look at the **validation.route.js** I raised, I find that the routes I added as complete strings give a much clearer overview of the route layout of the domain than the config approach.

Anyway, I'll change it if you're adamant! Let me know!

apps/web/src/server/views/app/layouts/README.md
brpins: I think the _ makes sense if they coexist in the same folder as the pages, but in being in their own layouts folder, the _ is redundant in my opinion. The visual separation is now the fact they're in /layouts

apps/web/src/styles/5-elements/_elements.page.scss
brpins: Ok... as an observation, I think we have too many folders/files for sass right now. I think the structure lends itself more to a project where we're building all the CSS, but in relying on 99% of it from GDS, it felt overly-fragmented to me...

apps/web/src/styles/7-components/_table.scss
brpins: I think editorconfig is only applying to existing files, and if I create a new file, it's defaulting to my VScode preferences

apps/web/src/server/app/app.express.js
brpins: Line 95? That's an eslint requirement to make sure we don't start adding .prototype values as filters

apps/web/src/server/app/validation/__dev__/local-planning-departments.js
brpins: Yes

apps/web/src/server/app/inspector/inspector.router.js
brpins: Ok, I did wonder why though as it exports a router... It would have been my third choice after *.router.js and *.routes.js !

apps/web/src/styles/5-elements/_elements.page.scss
radum: It is :) I was expecting way more custom stuff to begin with and I planned for that. But I think as we move fwd long term from what I have seen we will build more custom stuff and it will make sense. Refactoring later will be harder IMHO than using it like this now.

apps/web/src/styles/7-components/_table.scss
radum: Hmm, it shouldn't, it should work for the entire workspace.

brpins: I checked, and this is a tab. This is kinda the problem with tabs, in that they aren't predictable. If I re-apply prettier, it looks the same... Are you a tabs guy then? :(

apps/web/src/server/app/inspector/inspector.router.js
radum: I don't mind either one of them. I use route out of memory muscle. If you refactor all of them in a separate PR and tell us to rebase I will 👍🏻 .

Probably .routes.js sounds better.

apps/web/src/client/components/accessible-autocomplete.js
radum: That is what it will happen, a different chunk will be split and the code will fetch it only if your IF executes.

apps/web/src/styles/7-components/_table.scss
radum: :) yeah

Tabs are more accessible than spaces because they represent indentation visually and non-visually. Tabs allow developers to choose the indentations that make the most sense for them. Tabs improve how code is interpreted in non-visual mediums.

also this:

> Just had a conversation about indentation with a blind programmer, and realized this is the most accessible indentation convention as well. Tabs as indentation means # tabs = indent level, and all spaces (including alignment spaces) can be simply ignored when reading out code.
— @fantasai (W3C CSSWG specifications writer, 2004 to present)

apps/web/src/styles/main.scss
radum: I merged a fix for this so you can follow that example.

radum (approved):

PR #232, opened 2022-04-06
BOCM-56

james-hill-kc:
## Describe your changes

Templates, controller functions, validation pipes etc. for case officer journey check and confirm page + complete page

https://pins-ds.atlassian.net/browse/BOCM-56

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/views/app/layouts/app-wide.layout.njk
+++ b/apps/web/src/server/views/app/layouts/app-wide.layout.njk
@@ -1,3 +1,33 @@
 {% extends "app/layouts/app.layout.njk" %}
 
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+
 {% set containerClasses = 'pins-width-container' %}
+
+{% block content %}
+	<div class="govuk-grid-row">
+		<div class="govuk-grid-column">
+			{% if errorSummary %}
+				{{
+					govukErrorSummary({
+						titleText: "There is a problem",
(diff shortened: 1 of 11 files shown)


apps/web/src/server/views/app/layouts/app-wide.layout.njk
```diff
@@ -1,3 +1,33 @@
 {% extends "app/layouts/app.layout.njk" %}
 
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+
 {% set containerClasses = 'pins-width-container' %}
+
+{% block content %}
```
brpins, line 7: I'm not sure why this change is needed – this extends from app.layout.njk, which is handling all of this anyway?

james-hill-kc: The `app.layout.njk` doesn't handle this, it's only in the derived templates such as `app-two-column.layout.njk`. Perhaps we should move it into the base layout instead?

brpins: Oh, right... I actually simplified the layouts in my new PR (this app-wide layout no longer exists) and it takes care of what you're doing here, so not sure if we should spend much time on this here. Yours should just work once both are merged and this file is removed.

radum (approved):

radum (approved):

PR #261, opened 2022-04-13
add transition state error handler

brpins:
## Describe your changes
Add explicit exception to failed state transitions, and accompanying middleware.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/app/middleware/error-handler.js
+++ b/apps/api/src/server/app/middleware/error-handler.js
@@ -1,11 +1,11 @@
+import { TransitionStateError } from '../state-machine/household-appeal.machine.js';
+
 /**
- * @param {Error} error error that requires handling
- * @param {Request} _request API request
- * @param {Response} response API response
- * @param {Function} next next
- * @returns {Response} API Response
+ * The default catch-all error handler.
+ *
+ * @type {import('express').ErrorRequestHandler}
  */
-function errorHandler(error, _request, response, next) {
+export function defaultErrorHandler(error, _request, response, next) {
 	if (response.headersSent) {
 		return next(error);
 	}
@@ -15,4 +15,19 @@ function errorHandler(error, _request, response, next) {
 	return response.send({ error: error.message });
 }
 
-export default errorHandler;
+/**
+ * Handle any requests thrown by failed state transitions within the state machine.
+ *
+ * @type {import('express').ErrorRequestHandler}
+ */
+export const stateMachineErrorHandler = (error, _, response, next) => {
+	if (error instanceof TransitionStateError) {
+		response.status(409).send({
+			errors: {
+				appeal: error.message
+			}
+		});
+	} else {
+		next();
+	}
+};
--- a/apps/api/src/server/app/middleware/error-handler.js
+++ b/apps/api/src/server/app/middleware/error-handler.js
@@ -28,6 +28,6 @@ export const stateMachineErrorHandler = (error, _, response, next) => {
 			}
 		});
 	} else {
-		next();
+		next(error);
 	}
 };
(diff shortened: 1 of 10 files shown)


apps/api/src/server/app/middleware/error-handler.js
```diff
@@ -15,4 +15,19 @@ function errorHandler(error, _request, response, next) {
...
+export const stateMachineErrorHandler = (error, _, response, next) => {
+	if (error instanceof TransitionStateError) {
+		response.status(409).send({
+			errors: {
+				appeal: error.message
+			}
+		});
+	} else {
+		next();
+	}
```
mayya-bond-kc, line 32: does there need to be an `else`? I would have thought that 
```
if (...) {
    response.send() 
}
next()
``` 
would produce the same results?

apps/api/src/server/app/state-machine/household-appeal.machine.js
```diff
@@ -3,31 +3,59 @@ import { validationStates, validationActions } from './validation-states.js';
...
 
-const transitionState = function(context, status, machineAction) {
+const transitionState = function (context, status, machineAction) {
 	const service = interpret(createHouseholpAppealMachine(context));
 	service.start(status);
 	service.send({ type: machineAction });
 	const nextState = service.state;
 	service.stop();
+
+	if (nextState.value === status) {
```
mayya-bond-kc, line 53: Not sure that this is the best way to check if there has been a transition. What if we have an instance when the action loops us back to the initial state of the appeal?

mayya-bond-kc (dismissed):
Looks fine, the comments are non-breaking, but something to consider

apps/api/src/server/app/middleware/error-handler.js
brpins: no... response would get sent, but then the code would continue and `next` would get called and the middleware chain would continue, likely to the final error handler which then attempts to send a 500, and we'd get "Headers already sent" error or something.

Basically, we should never call next if this is our final action of the request

apps/api/src/server/app/state-machine/household-appeal.machine.js
brpins: There's a use case where we could "advance" a state to itself? Mmmm, ok... well, I don't know this state machine library at all – is there some way to check it did something valid?

apps/api/src/server/app/middleware/error-handler.js
mayya-bond-kc: so would using `return response.send()` do what i think it will?

apps/api/src/server/app/state-machine/household-appeal.machine.js
mayya-bond-kc: yep, we check if the transition was valid in the household appeal state machine tests. So we can check if the `nextState.changed` is their `false` or `undefined` and that would tell us that the transition did not go through

apps/api/src/server/app/middleware/error-handler.js
brpins: Return `response.send()` is another way to do this yes, but our (stupid) eslint rules then see that there's a return and want you to add returns for all if conditions, so you'd have to write:

```
if (...) {
    return response.send() 
}
return next();
```
IMO, that's uglier than an if/else, and somewhat inverts the assumed return type of {void}.

apps/api/src/server/app/state-machine/household-appeal.machine.js
brpins: Ok, let me change this to use `.changed`

brpins: Added, and also for now, had to add an opt-in flag to the transition state throwing an error as it breaks some tests in validation

mayya-bond-kc (approved):

PR #272, opened 2022-04-19
Feature/bocm 358 household appeal state machine

mayya-bond-kc:
## Describe your changes

refactors backend so that an appeal can have multiple states

## Type of change 🧩

- [ ] Documentation update


apps/api/src/server/app/inspector/appeal-formatter.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'lodash';
```
radum, line 1: I think should be lodash_es. And we should refactor all of them that are still using lodash.

mayya-bond-kc: how come? i saw Ben using lodash_es as well

apps/api/src/server/app/inspector/inspector.controller.js
```diff
@@ -1,49 +1,9 @@
 import _ from 'lodash';
```
radum, line 1: Same here it should be lodash-es

apps/api/src/server/app/inspector/appeal-formatter.js
radum: `lodash-es` is the esm version and even if the cjs works it is better to use the native version. Also the FE might use it in the future and it will help with bundle size.

radum (approved):

PR #279, opened 2022-04-20
Bocm 49 user auth management

radum:
## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/views/auth/unauthorized.njk
```diff
@@ -0,0 +1,9 @@
+{% extends "app/layouts/app-two-column.layout.njk" %}
+
```
brpins, line 2: You need to define the pageTitle, else the browser tab will just have something generic unrelated to the page.
`{% set pageTitle = 'Sorry, there is a problem with your login' %}`

By setting a page title, the page layout will automatically introduce that same page title into an <h1> element, which as per your template below, will result in double headings. You can either drop the h1 in your `pageContent`, or override the inserted heading using the `pageHeading` block.

apps/web/src/server/app/auth/auth.controller.js
```diff
@@ -1 +1,29 @@
-/* eslint-disable-line unicorn/no-empty-file */
+// @ts-ignore
```
brpins, line 1: Why do we need this?

apps/web/src/server/app/auth/auth.controller.js
```diff
@@ -1 +1,29 @@
...
+import { ssoConfig } from '../../config/sso.js';
+import { cca } from '../../lib/sso.js';
+
+/**
+ * Path to redirect to after successful login.
+ *
+ * @type {import('@pins/express').QueryHandler<{}>}
```
brpins, line 8: I think it would be useful to define the query params, as I'm not sure it's clear what "code" is...
```
/**
 * @typedef {object} AuthRedirectQuery
 * @property {string} code - <What this is>
 */
```
and then you can plug it in via:
```
/** @type {import('express').RequestHandler<?, ?, ?, AuthRedirectQuery>} */
```
(You could use the @pins/express handler but it's primarily for handling routes rendering html)

apps/web/src/server/config/sso.js
```diff
@@ -0,0 +1,13 @@
+import { loadEnvironment } from '@pins/platform';
+
+loadEnvironment(process.env.NODE_ENV);
```
brpins, line 3: isn't the idea that loadEnvironment is only used within apps/web/src/service/config/config.js, and then everything else imports from there?

jsconfig.json
```diff
@@ -5,7 +5,7 @@
     "checkJs": true,
     "esModuleInterop": true,
     "moduleResolution": "node",
-    "module": "ES2020",
+    "module": "es2022",
```
brpins, line 8: No objections, just curious... What's the ES2022-specific introduction?

apps/web/src/server/lib/sso.js
```diff
@@ -0,0 +1,58 @@
...
+				// eslint-disable-next-line unicorn/explicit-length-check
+				if (rule.groups.filter((element) => creds.includes(element)).length < 1) {
+					console.error('User does not have this group');
+					return false;
+				}
+				break;
+
+			case 'roles':
+				// eslint-disable-next-line unicorn/explicit-length-check
+				if (rule.roles.filter((element) => creds.includes(element)).length < 1) {
```
brpins, line 43: (same as above)

apps/web/src/server/lib/sso.js
```diff
@@ -0,0 +1,58 @@
...
+			logLevel: msal.LogLevel.Verbose
+		}
+	}
+});
+
+/**
+ * Checks if the request passes a given access rule
+ *
+ * @param {string} method HTTP method for this route
+ * @param {Object} rule access rule for this route
```
brpins, line 25: Can you add a TODO here for tightening these `Object` and `Array` types?

apps/web/src/server/app/auth/auth.router.js
```diff
@@ -0,0 +1,29 @@
...
+router.route('/redirect').post(handleRedirect);
+
+// TODO: This should actually be run for each main route, not just `/auth`.
+router.use((request, response, next) => {
+	if (!request.session) {
+		throw new Error('No session found for this request'); // TODO: Handle this gracefully
+	}
+
+	// Add session nonce for crsf
+	request.session.nonce = cryptoProvider.createNewGuid();
```
brpins, line 20: How is this being used to prevent CSRF? Won't we have to start sending this from the UI with every post/put/patch?

apps/web/src/server/app/auth/auth.guards.js
```diff
@@ -0,0 +1,75 @@
...
+	} else {
+		console.error('No session found for this request');
+		response.redirect('/auth/unauthorized');
+	}
+}
+
+export function hasAccess(options) {
+	// eslint-disable-next-line consistent-return
+	return (request, response, next) => {
+		if (request.session) {
```
brpins, line 26: can you add a todo about introducing a session service to encapsulate auth session logic? Thanks!

jsconfig.json
brpins, line 8: not an issue, just curious – what's introduced that requires this to be es2022?

radum: Dynamic imports with top level await.

```js
const { authConfig } = await import('./components/auth.js');
```

That would have been flagged without es2022.

apps/web/src/server/app/auth/auth.router.js
radum: When you login there is a series of redirect going from our app to Azure, you can argue that when it comes back if its not the same session someone captured that redirect urls from Azure and they could authenticate and get the tokens from it.

I followed their best practices on how this can be handled. To be honest didn't read to much into the attack vectors examples for this particular thing but I imagine is just that as it makes sense from a possible issue.

jsconfig.json
radum: Dynamic imports with top level await.

```
const { authConfig } = await import('./components/auth.js');
```

That would have been flagged without es2022.

PR #289, opened 2022-04-21
Feature/bocm 351 endpoint for case officer to upload statement

mayya-bond-kc:
## Describe your changes

- creates endpoints to submit statements and final comments
- when appeal has more than 1 status, updates only the necessary statuses

## Type of change 🧩

- [x] Documentation update


apps/api/src/server/app/middleware/validate-file-upload.js
```diff
@@ -0,0 +1,20 @@
+import { composeMiddleware, mapMulterErrorToValidationError } from '@pins/express';
+import multer from 'multer';
+import { body } from 'express-validator';
+import { handleValidationError } from './handle-validation-error.js';
+
+export const validateFileUpload = function(filename) {
```
brpins, line 6: I don't think we should abstract this:  Depending on the endpoint, the .single() method will sometimes be .fields() or .array() (which in turn changes the custom validator passed to body), and the storage mechanism, limits, and message will differ. Multer actually transforms the FormData into a body that express-validator can use, and by adding handleValidationError immediately after checking the file means that the response will be sent before any other fields in body can be validated.

We tried to do it this way on web and ultimately only ended up keeping the different storages as singletons and importing them.

apps/api/src/server/app/middleware/handle-validation-error.js
```diff
@@ -0,0 +1,17 @@
+import { validationResult } from 'express-validator';
```
brpins, line 1: I also extracted this in my BOCM-429 commit!

apps/api/src/server/app/middleware/validate-file-upload.js
brpins: Having checked, according to the UX, the user can upload multiples files, so you would actually need .array() here and a different custom validator. Check `validateAppealDocuments` in  apps/web/src/server/app/validation/validation.pipes.js for an example.

PR #296, opened 2022-04-22
Update azure-pipelines-build.yml for Azure Pipelines

mayya-bond-kc:


andylangridge (dismissed):

andylangridge:
You'll probable have to add this to the web pipeline too to get that passing

andylangridge (approved):

PR #305, opened 2022-04-27
add integration testing setup/tests for web

brpins:
## Describe your changes
- Add factories and fixtures for creating test data
- Add jest setup
- Add integration tests for /validation domain

## Type of change 🧩

- [ ] Documentation update


packages/platform/testing/util.js
```diff
@@ -0,0 +1,35 @@
...
+		id += 1;
+		return id;
+	};
+})();
+
+/**
+ * @param {number | string} value
+ * @returns {string}
+ */
+export const pad = (value) => {
```
radum, line 31: String has padStart these days, no more drama around this. So you might want to use that.

packages/platform/testing/html-parser.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Parses HTML and returns a root element.
+ *
+ * @param {string} html
+ * @param {Partial<import('node-html-parser').Options> & { rootElement?: string }} [options={}]
+ * @returns {HTMLElement}
+ */
+export const parseHtml = (html, { rootElement = 'main', ...options } = {}) => {
+	const prettifiedHtml = prettyPrint(
```
radum, line 12: Is `prettyPrint` required for nice diffs?

packages/platform/testing/util.js
brpins: oh, didn't realise! Will change

packages/platform/testing/html-parser.js
brpins: nicer snapshots... nunjucks can end up inserting a lot of whitespace because of the way it interprets it's formatting, so I'm using the parser to strip the whitespace entirely, and then pretty print to format it nicely for snapshotting

radum: I thought so.

apps/web/package.json
```diff
@@ -56,6 +57,9 @@
   "devDependencies": {
     "@babel/core": "^7.17.9",
     "@babel/preset-env": "^7.16.11",
+    "@faker-js/faker": "^6.2.0",
```
radum, line 60: Lock the version please.

packages/platform/testing/util.js
brpins: Done

packages/platform/testing/html-parser.js
brpins: I would say it's worth it even if it's slower, because the primary reason to do it is not so it looks nicer in the file but so that snapshots can be read by a developer much easier when asserting them or doing diffs...

apps/web/package.json
```diff
@@ -17,7 +17,7 @@
     "sass": "node ./scripts/compile-css.js",
     "watch:rollup": "chokidar \"src/client/**/*.js\" --silent -c \"npm run rollup\"",
     "watch:sass": "chokidar \"src/styles/**/*.scss\" --silent -c \"npm run sass\"",
-    "test": "echo 'TOOD: add test script'",
+    "test": "NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules npx jest",
```
radum, line 20: Use `cross-env` to set the env vars to make sure it won't crash on Windows or what not.

See above.

packages/platform/http/ttl.js
```diff
@@ -31,10 +31,10 @@ export function createTtlHandler() {
 				if (cachedResponse) {
 					// If a previously cached response has not yet expired, we can serve
 					// this without performing an http request
-					if (cachedResponse.expires > now + ttl) {
+					if (cachedResponse.expires > now) {
             options.context.servedFromCache = true;
```
radum, line 35: Can you do me a favour and format this fill again, as it has a mixture of spaces and tabs? Thank you.

apps/web/package.json
brpins: Done

apps/web/package.json
brpins: Done

packages/platform/http/ttl.js
brpins: Done

radum:
Should we move the testing folder from the `src/server` folder? I can't get off the feeling it doesn't belong there.

apps/web/package.json
```diff
@@ -56,6 +57,9 @@
   "devDependencies": {
     "@babel/core": "^7.17.9",
     "@babel/preset-env": "^7.16.11",
+    "@faker-js/faker": "6.2.0",
+    "@pins/api": "file:../api",
+    "@pins/web": "file:.",
```
radum, line 62: Why do we need this?

brpins: So, by default, npm workspaces will be symlinked via node_modules to allow for importing from them like normal. This is why we can import from @pins/platform and @pins/express etc.

However, because the web and api are called "web" and "api" respectively, these are symlinked into node-modules under those names, without the @pins namespace. In order to continue the convention of @pins/*, there was two options:

1. Rename the "web" and "api" package name to "@pins/web" and "@pins/api", or
2. Alias them so that they can retain their names but still be accessed by the @pins namespace.

I tried (1) first, but it meant that the npm scripts then needed the new names (i.e. npm run dev --workspace=@pins/web etc), which seemed less convenient. So, this second way creates a symlink by pointing to a relative folder.

As to why do this at all, it's to avoid having to form long relative path names from the testing folder that are pathed across packages. If we want to use a function/type from the api for testing purposes, we can treat the api like a package and share things via an index.js in its root.

radum: I see. I still want to do the first option. Adding @pins/ is not that big of a deal, and can be fixed by simply running `npm run dev` in root, and turbo will take care of the entire thing anyways running both of them with log grouping and all that stuff.

Adding the index file and the reference like this is more confusing than it help IMHO. So please rename the packages.

brpins: Ok – are we sure this isn't going to break anything in the pipelines by renaming the app packages like this?

brpins: I've committed this change.

radum:
Can you also please update the readme file

```
# OR you can manually run them
npm run dev --workspace=api
npm run dev --workspace=web
```

this should say @pins/api and web.

apps/web/index.js
```diff
@@ -0,0 +1 @@
+export { app } from './src/server/app/app.express.js';
```
radum, line 1: We don't need this file anymore right?

apps/web/package.json
```diff
@@ -17,7 +17,7 @@
     "sass": "node ./scripts/compile-css.js",
     "watch:rollup": "chokidar \"src/client/**/*.js\" --silent -c \"npm run rollup\"",
     "watch:sass": "chokidar \"src/styles/**/*.scss\" --silent -c \"npm run sass\"",
-    "test": "echo 'TOOD: add test script'",
+    "test": "cross-env NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules npx jest",
```
radum, line 20: Using npx will imply latest version. Might want to keep it using the local bin, so drop the npx.

apps/web/src/server/app/validation/validation.controller.js
```diff
@@ -426,6 +427,8 @@ export async function confirmReviewOutcome({ params, session }, response) {
 	);
 	const appeal = await validationService.findAppealById(params.appealId);
 
+console.log({ reviewOutcome });
```
radum, line 430: Need this?

apps/web/src/server/app/validation/validation.service.js
```diff
@@ -33,7 +33,7 @@ export function findAllAppeals() {
  * @returns {Promise<Appeal>}
  */
 export function findAppealById(appealId) {
-	return get(`validation/${appealId}`, { context: { ttl: 1000 } });
+	return get(`validation/${appealId}`, { context: { ttl: 10_000 } });
```
radum, line 36: :) Nice usage of _

apps/web/package.json
```diff
@@ -27,6 +27,7 @@
     "@azure/msal-node": "1.8.0",
     "@hapi/boom": "9.1.4",
     "accessible-autocomplete": "2.0.4",
+    "babel-jest": "^27.5.1",
```
radum, line 30: Do we need this though? Is it for using ESM within the test files?

apps/web/index.js
brpins: Removed

apps/web/package.json
brpins: What do you mean by imply latest version? npx will only source the latest package from npm if it it's not present in your package.json dependencies (like, if you wanna run jest without installing it). Running without npx wouldn't guarantee that the local binary is used.

apps/web/src/server/app/validation/validation.controller.js
brpins: Definitely. It's vital.

apps/web/package.json
brpins: Yes, for exactly that.

apps/web/src/server/app/app.express.js
```diff
@@ -66,13 +66,15 @@ if (config.authDisabled) {
...
 
 // CSRF middleware via session
-app.use(
-	csurf({ cookie: false }),
-	(request, response, next) => {
-		response.locals.csrfToken = request.csrfToken();
-		next();
-	}
-);
+if (process.env.NODE_ENV !== 'test') {
```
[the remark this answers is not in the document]
brpins: we don't actually have it... I can add it, but isn't it a somewhat redundant alias if the node env is available ubiquitously anyway?

apps/web/src/server/app/validation/validation.controller.js
brpins: Removed!

radum (approved):

PR #336, opened 2022-05-05
add integration tests for auth

brpins:
## Describe your changes

Auth tests!

## Type of change 🧩

- [ ] Documentation update


apps/web/.env.test
```diff
@@ -31,18 +31,12 @@ API_HOST=http://test
...
-AUTH_CLIENT_ID=<value goes here>
-AUTH_CLOUD_INSTANCE_ID=<value goes here>
-AUTH_TENANT_ID=<value goes here>
-AUTH_CLIENT_SECRET=<value goes here>
-AUTH_REDIRECT_URI=<value goes here>
+AUTH_CLIENT_ID=auth_client_id
+AUTH_CLOUD_INSTANCE_ID=auth_cloud_instance_id
+AUTH_TENANT_ID=auth_tenant_id
+AUTH_CLIENT_SECRET=auth_client_secret
+AUTH_REDIRECT_URI=auth_redirect_uri
```
radum, line 38: This should be `/auth/redirect` even in test as that is how the Azure app has been configured and the route on our side is like that.

apps/web/src/server/app/auth/auth.controller.js
```diff
@@ -97,7 +97,7 @@ export function authSignIn(request, response, next) {
  * @param {import('express').NextFunction} next express next function
  */
 export async function handleRedirect(request, response, next) {
-	if (request.query.state) {
+	try {
```
radum, line 100: The only other code that could benefit from this is `cryptoUtils`. I am not expecting that to fail, but having the state is mandatory.

Why did you change the if to a try?

apps/web/testing/mocks/auth.js
```diff
@@ -0,0 +1,27 @@
+import { installSessionMock } from './session.js';
+
+/**
+ * Install an authenticated user account to the session with membership to the
```
radum, line 4: Pretty please add more docs says that the object structure (idTokenClaims and all that) is coming from Azure MSAL and not set by us. It helps putting into context the structure and naming.

apps/web/.env.test
brpins: Ok... but should this even be an env variable then if it's fixed? I see there is `redirectUri: process.env.AUTH_REDIRECT_URI || '/auth/redirect',` in the config setup, which suggests it can be overriden?

radum: During testing or other env changes it will come handy to have it as an ENV as we can change it without building and deploying the entire solution again. Its one of those that stay like that long term but should have the option to change it on the fly if need be and skip a build step.

apps/web/src/server/app/auth/auth.controller.js
brpins: This was actually a bug fix. Sending a rogue state value like `x` to this endpoint was actually causing the express app to hang somehow (the test would time out).

Furthermore, it's not sufficient just to check that there is a state in the query, as parsing/decrypting this state also relies on `request.session.key`, which could be undefined. Given the multiple possibilities for failure, and an existing try/catch being there, I just moved it to the top level, to catch any issue.

apps/web/testing/mocks/auth.js
brpins: Ok – I think we can describe this by introducing an extended jsdoc type definition for azure's `AccountInfo` type, tailored to the PINS response.

apps/web/src/server/app/auth/auth.controller.js
radum: Not sure how one could ever send a rogue state? You mean from our code?

As for that key dependency all of it is encrypted in `authSignIn(request, response, next)` so in theory you should have everything you need.

What I am trying to say is that if I hit the redirect url directly, what happens with the controller if we don't check for the presence of state?

Also maybe state is not the best name for it as its specific to auth.

apps/web/.env.test
brpins: If we change the .env.test to use `/auth/redirect` then that wouldn't prove that the environment variable is being picked up, as it'd be identical to the default value. If we have the option to change it, then the test needs to be able to identify between two possible values...

apps/web/src/server/app/auth/auth.controller.js
brpins: This function can be hit directly from a urlL Like you say, if you hit the url directly, then anything can be sent as state. Even if azure hits the url, the session may not longer exist, which also causes this handler to throw.

We can revert to an `if` but then we also have to add an `if` to the session key, and session nonce. Basically, there's a number of points of failure if this endpoint is hit improperly (or even properly, if the session has disappeared), so a try catch seems safer?

radum: I see what you mean now. The crypto decrypt could fire a TypeError if its deps are missing or rubish. Not sure if that can ever happen but I agree now that is safer like this.

radum (approved):

radum (approved):

PR #344, opened 2022-05-12
Feature/bocm 486 create file storage service

mayya-bond-kc:
## Describe your changes

- creates new service for azure blob storage integration

PS I know this is missing some documentation and some endpoints (Dockerisation etc), but thought it would be better to start moving this into the develop branch sooner rather than later

## Type of change 🧩

- [ ] Documentation update


apps/document-storage/.env.development
```diff
@@ -0,0 +1,23 @@
...
+# Default : 3443
+HTTPS_PORT=3443
+
+# - - - - - - - - - Swagger - - - - - - - 
+
+SWAGGER_JSON_DIR=./src/server/swagger-output.json
+
+# - - - - - - - Blob Storage - - - - - -
+
+AZURE_BLOB_STORE_CONNECTION_STRING=DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;
```
radum, line 22: This key should not be here.

radum: Any secret should be set in the gitignored file. Here is just for reference.

mayya-bond-kc: This secret key actually not secret at all. This is the same secret key that everyone gets when using the local version of the azure blob storage. Here is a public page with all this info: https://docs.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string

In the dev/test envs, we will just need to override this connection string.

francescobrenci-AR (approved):

PR #345, opened 2022-05-13
add logging via pino to apps/web

brpins:
## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/lib/logger.js
```diff
@@ -0,0 +1,35 @@
+import config from '@pins/web/environment/config.js';
+import path from 'node:path';
+import pino from 'pino';
+
+/** @type {import('pino').TransportTargetOptions[]} */
+const targets = [];
+
+if (config.isProd) {
+	targets.push({
```
radum, line 9: I think we need the pretty version here also, but maybe with a different level. Just to make sure the logstream has the info we need when we look at it.

apps/web/src/server/lib/logger.js
```diff
@@ -0,0 +1,35 @@
...
+		target: 'pino/file',
+		level: 'trace',
```
radum, line 11: Shouldn't this come from the env vars?

apps/web/src/server/lib/logger.js
brpins: so you want to pretty-print before it's written into the log file? I'd have thought it's better to save them in a one-log-per-line for a log file? (bear in mind they can still be pretty printed from the log file if needed)

apps/web/src/server/lib/logger.js
brpins: It's not really necessary, as at the top level of the pino instantiation, the log level is passed from the `config`, which means logs won't be captured for anything lower level than the env var. Putting `trace` in the transport is basically saying "we're deferring to the master log level".

apps/web/src/server/lib/logger.js
radum: I would say debug for file and info for console. But I think both should be configured separately.

radum: Good point, I wonder if we should replace morgan with https://github.com/pinojs/pino-http to be consistent with the tool and pretty that one to console only for now.

brpins: Have now updated the PR with a change to transport construction. Regarding pino-http, I'll try it.

radum:
LGTM

brpins:
@radum can you re-approve? I had to resolve merge conflicts

PR #352, opened 2022-05-19
Add build pipeline that lints and tests entire repo

andylangridge:
## Describe your changes

Build pipeline that now runs linting and tests from the root of the repo. Previous build pipelines are still present until the linting issues reported by this new pipeline have been resolved.

The pipeline currently fails, but have tested that it runs end to end and it is reporting the issues we expect to see - https://dev.azure.com/planninginspectorate/back-office/_build?definitionId=126

https://pins-ds.atlassian.net/browse/BOAS-37

## Type of change 🧩

- [ ] Documentation update


azure-pipelines-build.yml
```diff
@@ -0,0 +1,114 @@
...
+              developmentBranch: develop
+              pathFilters:
+                - name: packages
+                  path: packages
+                - name: web
+                  path: apps/web
+                - name: api
+                  path: apps/api
+              stepName: check_files
+      - name: Build Web
```
brpins, line 60: I notice that on the jobs for this PR, the lint tasks have failed, but the build tasks ran anyway? Shouldn't they have aborted?

azure-pipelines-build.yml
```diff
@@ -0,0 +1,114 @@
...
+          - template: ../steps/azure_web_app_acr_push.yml@templates
+            parameters:
+              azurecrName: $(azurecrName)
+              dockerfilePath: $(dockerFilePath)
+              repository: $(repository)
+              workingDirectory: $(Build.Repository.LocalPath)
+              runTests: false
+        variables:
+          - template: apps/api/pipelines/azure-pipelines-variables.yml@self
+    validateName: Lint, Test, & Build Applications
```
brpins, line 114: I think this line has the effect of creating a build task reported as:
Planning-Inspectorate.back-office (Lint, Test, & Build Applications Run Linting & Tests) 

This is a bit of a mouthful – perhaps this `validateName` should be "Prebuild" or something, so we can achieve a shorter (Prebuild - Run Linting & Tests) ?

brpins:
Minor thing, but the new task prefix is "Planning-Inspectorate.back-office" – it is possible to align it with the build task naming convention (which uses "Back Office" prefix)?

azure-pipelines-build.yml
andylangridge: Which one are ya looking at? This run they skipped - https://dev.azure.com/planninginspectorate/back-office/_build/results?buildId=30429&view=results.

In the "condition" attribute I've included the `succeeded()` condition. So they should only run on success. It might be because you were looking at one I ran while testing I hope!

andylangridge:
> Minor thing, but the new task prefix is "Planning-Inspectorate.back-office" – it is possible to align it with the build task naming convention (which uses "Back Office" prefix)?

Yep! I just need to rename that in the GUI so no issues there :)

azure-pipelines-build.yml
andylangridge: This is the name of the "Stage". And since we only have 1 stage, this is actually completely irrelevant as it doesn't display in the GUI anywhere. In the templates there is a default value of "Validate" so I'll just use that and delete this line. It's only used for referencing within this template (i.e. if one stage depends on another). But like I said, we only have 1 stage (which includes 3 jobs) so the default should be fine.

PR #426, opened 2022-07-01
Feature/boas 92 be find a case by a case officer admin inspector

MariaValeroGonzalezKC:
## Type of change 🧩

- [√ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

- [√ ] I have performed a self-review of my own code
- [√] I have double checked this work does not include any hardcoded secrets or passwords
- [√ ] I have provided details on how I have tested my code
- [√ ] I have referenced the ticket number above
- [√ ] I have provided a description of the ticket
- [√ ] I have included unit tests to cover any testable code changes


apps/api/src/server/applications/search/__tests__/case-search.test.js
```diff
@@ -0,0 +1,172 @@
...
+			OR: [
+				{
+					title: { contains: 'EN010003 - NI Case 3 Name' }
+				},
+				{
+					reference: { contains: 'EN010003 - NI Case 3 Name' }
+				},
+				{
+					description: { contains: 'EN010003 - NI Case 3 Name' }
+				}
```
mayya-bond-kc, line 70: We can probably generalise this string`'EN010003 - NI Case 3 Name'` across the file and put it in a variable

apps/api/src/server/applications/search/__tests__/case-search.test.js
```diff
@@ -0,0 +1,172 @@
...
+			]
+		}
+	})
+	.returns(applicationsCount);
+
+test('gets applications using search criteria', async (t) => {
```
mayya-bond-kc, line 76: Some more test cases to cover:
- What if there are no results to the query?
- What if someone requests `pageNumber: 0` or `pageSize: 0`
- Explicit tests for the different roles which should be allowed to call this endpoint

apps/api/src/server/applications/search/case-search.controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+	const pageInfo = {
+		page: skipValue / resultsPerPage + 1,
+		pageSize: applications.length,
+		pageCount: Math.ceil(applicationsCount / resultsPerPage),
+		itemCount: applicationsCount,
+		items: mapApplicationsWithSearchCriteria(applications)
+	};
+
+	return response.send(pageInfo);
```
mayya-bond-kc, line 49: There is way too much going on in this controller. As we have done in other domains, let's use a service here that we call given the parameters sent to the endpoint.

apps/api/src/server/applications/search/case-search.routes.js
```diff
@@ -0,0 +1,50 @@
...
+			example: 'EN01',
+            type: 'string',
+            required: true
+        }
+		#swagger.parameters['role'] = {
+            in: 'body',
+			description: 'a valid role',
+            type: 'string',
+            required: true
+        }
```
mayya-bond-kc, line 26: we should be able to use an enum here to list which roles are allowed to call this endpoint

apps/api/src/server/applications/search/search.validators.js
```diff
@@ -0,0 +1,66 @@
...
+		});
+	} else if (!validPageSize(request.body.pageSize)) {
+		response.status(400).send({
+			errors: {
+				status: 'Page Size not in valid range'
+			}
+		});
+	} else {
+		next();
+	}
```
mayya-bond-kc, line 35: Firstly, I would say that the role should be the first thing that is validated. Second, why are we not using `express-validator` here?

apps/api/src/server/repositories/application.repository.js
```diff
@@ -18,3 +18,63 @@ export const getByStatus = (status) => {
...
+				}
+			]
+		},
+		include: {
+			subSector: {
+				include: {
+					sector: true
+				}
+			}
+		}
```
mayya-bond-kc, line 56: Do we need to return the sub-sector and sector information?

mayya-bond-kc (changes requested):
Looks good overall. My main concern is that the controller has too much going on there and that there are some test scenarios missing.

apps/api/src/server/applications/search/__tests__/case-search.test.js
MariaValeroGonzalezKC: Solved with feat(api/applications): searchString for testing (BOAS-92)

apps/api/src/server/applications/search/case-search.routes.js
MariaValeroGonzalezKC: After a bit of research, it seems like the way of using a sort of enum in JS would be by freezing the object, like adding Object.freeze(validRoles) after its declaration

apps/api/src/server/repositories/application.repository.js
lutamKC: we don't know yet what information we're going to need on the search results page, but if it's any similar to the dashboard we might need to sort the cases by sector

MariaValeroGonzalezKC: As we do not have a design saying the opposite, we are leaving it

apps/api/src/server/applications/search/__tests__/case-search.test.js
MariaValeroGonzalezKC: We checked with Ludovico, and he is happy with the results it yields

apps/api/src/server/utils/mapping/map-application.js
```diff
@@ -4,13 +4,11 @@ import { mapKeysUsingObject } from '../../utils/mapping/map-keys-using-object.js
...
 /**
- *
- * @typedef {{id: number, modifiedDate: number, reference: string}} ApplicationResponse
+ * @typedef {{id: number, reference: string, title:string, description: string, status: string, modifiedDate: number, publishedDate?}} ApplicationResponse
  */
 
 /**
- *
- * @param {import('@pins/api').Schema.Application} application
+ * @param {import('@pins/api').Schema.Application[]} application
```
DavidJWhiteDev, line 11: @MariaValeroGonzalezKC I forgot to undo the extra [] I added here, line should read:

* @param {import('@pins/api').Schema.Application} application

not 

* @param {import('@pins/api').Schema.Application[]} application

DavidJWhiteDev (changes requested):

DavidJWhiteDev (approved):

mayya-bond-kc (approved):

PR #461, opened 2022-07-18
Feature/boas 129 api endpoint to create update application

mayya-bond-kc:
## Describe your changes

- endpoint to create new case
- endpoint to update existing case

## Type of change 🧩

- [x] Documentation update


apps/api/prisma/migrations/20220712130328_create_grid_reference_and_service_customer/migration.sql
```diff
@@ -0,0 +1,56 @@
...
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[ApplicationDetails] ADD [locationDescription] NVARCHAR(1000),
+[mapZoomLevel] NVARCHAR(1000),
+[firstNotifiedAt] DATETIME2,
+[submissionAt] DATETIME2;
+
+-- CreateTable
+CREATE TABLE [dbo].[ServiceCustomer] (
```
lutamKC, line 12: so the technical name of the applicant is ServiceCustomer

apps/api/src/server/applications/application/application.validators.js
```diff
@@ -0,0 +1,154 @@
...
+ */
+const timestampToDate = (value) => {
+	return new Date(value);
+};
+
+export const validateCreateUpdateApplication = composeMiddleware(
+	body('geographicalInformation.gridReference.easting')
+		.isLength({ min: 6, max: 6 })
+		.toInt()
+		.withMessage('Easting must be integer with 6 digits')
```
lutamKC, line 91: are these messages meant to be shown to the user or they're just for our logs?

apps/api/src/server/applications/application/application.controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @type {import('express').RequestHandler}
+ */
+export const createApplication = async (request, response) => {
+	const mappedApplicationDetails = mapCreateApplicationRequestToRepository(request.body);
+
+	const application = await caseRepository.createApplication(mappedApplicationDetails);
+
+	const applicantIds = getServiceCustomerIds(application.serviceCustomer);
```
lutamKC, line 24: so this means that you're sending me back an array of applicant ids and (at least for now) I'll just have to pick the first one and send it back, right?

lutamKC (approved):

lutamKC:
@mayya-bond-kc, what do you think to do now with the branches? are you going to create the GET endpoint on a sub-branch of boas-166?

apps/api/prisma/migrations/20220712130328_create_grid_reference_and_service_customer/migration.sql
mayya-bond-kc: Exactly. This will include applicants and anyone else for other cases (not just applications)

apps/api/src/server/applications/application/application.validators.js
mayya-bond-kc: These messages will be in the body of the response but not in out logs (as this is the expected behaviour.

apps/api/src/server/applications/application/application.controller.js
mayya-bond-kc: Yep

PR #477, opened 2022-07-27
Feature/boas 165 start case

mayya-bond-kc:
## Describe your changes

- creates endpoint to start case

## Type of change 🧩

- [ ] Documentation update


apps/api/index.d.ts
```diff
@@ -1,3 +1,5 @@
 import * as Schema from './prisma/schema';
 
 export { Schema };
+
+export type CaseType = 'household' | 'full planning' | 'application';
```
DavidJWhiteDev, line 5: hmm I have seen this a few places, but thought I should ask it now.  The type of appeal is a "Householder Appeal" not "Household Appeal" - so should this say householder, not household?

DavidJWhiteDev (approved):

francescobrenci-AR (approved):

apps/api/index.d.ts
francescobrenci-AR: does this need confirmation

mayya-bond-kc: That's interesting :D no one said anything before. I am a little on the fence about making changes to the appeal side of the service

apps/api/src/server/applications/application/__tests__/start-case.test.js
```diff
@@ -0,0 +1,130 @@
...
+		}
+	});
+});
+
+test('throws an error if the application does not have all the required information to start', async (t) => {
+	const response = await request.post('/applications/3/start');
+
+	t.is(response.status, 400);
+	t.deepEqual(response.body, {
+		error: {
```
lutamKC, line 111: is it not 'errorS'?

apps/api/index.d.ts
francescobrenci-AR: cool - let's leave it as is, when we work on Appeals again and this is raise as a requirement it can be changed

apps/api/src/server/applications/application/__tests__/start-case.test.js
mayya-bond-kc: So at the minute, the default error handler returns `error` not `errors`. And I was not going to change it until I was sure that it was ok on the Web side. Do you know if it might be an issue?

apps/api/src/server/utils/application-factory-for-tests.js
```diff
@@ -9,39 +9,67 @@ const generateApplicationReference = () => {
...
+ *  description?: string | null,
+ * 	regions?: object[],
+ *  subSectorId?: number | null,
+ * 	subSector?: import('@pins/api').Schema.SubSector,
+ * 	zoomLevelId?: number | null
+ * }} arg
  * @returns {import('@pins/api').Schema.Case}
  */
-export const applicationFactoryForTests = ({ id, status = 'open', modifiedAt = new Date() }) => {
+export const applicationFactoryForTests = ({
```
lutamKC, line 25: would be nice to have common factories for BE and FE

apps/api/src/server/applications/case-admin-officer/case-admin-officer.routes.js
```diff
@@ -1,8 +1,8 @@
-import express from 'express';
+import { Router as createRouter } from 'express';
 import { asyncHandler } from '../../middleware/async-handler.js';
 import { getApplications } from './case-admin-officer.controller.js';
 
-const router = new express.Router();
```
lutamKC, line 5: is there a reason for all these? is it related to "that problem" with the routes?

apps/api/src/server/utils/application-factory-for-tests.js
mayya-bond-kc: We can do!

apps/api/src/server/applications/application/__tests__/start-case.test.js
lutamKC: it's not an issue as long as it's always 'error'. I can see a few lines up and and a few lines down you're returning errors

apps/api/src/server/utils/application-factory-for-tests.js
mayya-bond-kc: Except it is one thing to have a factory for the Database, and another to have a factory for the API responses for example. Not sure where you are using factories?

apps/api/src/server/applications/case-admin-officer/case-admin-officer.routes.js
mayya-bond-kc: Its that JSdoc issue. I was copying over what was done on Web

apps/api/src/server/applications/application/__tests__/start-case.test.js
```diff
@@ -0,0 +1,130 @@
...
+			subSector: 'Missing subSector',
+			title: 'Missing title'
+		}
+	});
+});
+
+test('throws an error if the application is not in draft state', async (t) => {
+	const response = await request.post('/applications/4/start');
+
+	t.is(response.status, 409);
```
lutamKC, line 124: who knew there was a 409?? thank you for this

lutamKC (approved):

apps/api/src/server/utils/application-factory-for-tests.js
lutamKC: for unit tests

lutamKC: happy to have a chat about this at some point. absolutely not urgent

apps/api/src/server/applications/application/__tests__/start-case.test.js
mayya-bond-kc: Any endpoint that has a state transition, if that transition fails, then we should see a 409

apps/api/src/server/utils/mapping/map-case-status-string.js
```diff
@@ -0,0 +1,14 @@
+import { get } from 'lodash-es';
+
+/**
+ *
+ * @param {string} caseStatus
+ * @returns {string}
+ */
+export const mapCaseStatusString = (caseStatus) => {
+	const caseStatusMap = {
+		pre_application: 'Pre-Application'
```
lutamKC, line 10: I was wondering whether the format of the statuses names would become a problem at some point.
maybe we want CaseStatus to have name, displayEn and displayCy?

apps/api/src/server/applications/application/__tests__/start-case.test.js
mayya-bond-kc: So validation error handler returns `errors`, hence the difference. i can make it so it is consistent?

apps/api/src/server/utils/mapping/map-case-status-string.js
mayya-bond-kc: hmmmm good question. TBH i dont think Welsh is in plans for the MVP. @francescobrenci-AR ?

lutamKC: oh, well, even without Welsh, isn't it better to have a `name` for being used in the code a displayName for the template?

mayya-bond-kc: sorry don't quite get what you mean here @lutamKC

lutamKC: oh sorry I forgot an "AND" and the sentence did not make sense.

i was just saying that since we're saving the status name as a regular text (with spaces, hyphens and stuff), we need things like this map above to use it in the code. so I was asking if it's worth to create an extra field in the caseStatus table that contains a property called `name` in which we put the "code-safe" name of the status.
which is exactly what we do with regions, zoom levels, sectors etc..

apps/api/src/server/applications/application/__tests__/start-case.test.js
lutamKC: yes, that would be better

apps/api/src/server/utils/mapping/map-case-status-string.js
mayya-bond-kc: Ah yes, so my intention is only to store the code-safe value of the status. The mapping will then post likely come from the code itself

francescobrenci-AR: @mayya-bond-kc Welsh translation is not something we are doing for MVP but is happening at some point after that - I don't have a timeframe or requirements though. It's very much a WIP still

apps/api/src/server/applications/application/__tests__/start-case.test.js
mayya-bond-kc: done

PR #508, opened 2022-08-24
fix(web/applications): key dates future date validation (BOAS-315)

DavidJWhiteDev:
## Describe your changes

Fix the validation of future date in the Key dates page.  Issue was that the Date() function month parameter is 0-11, was passing in the real month number 1-12 - hence past dates in this month were considered future - next month.

## BOAS-315 - Key Dates page - Future Date validation not working
     https://pins-ds.atlassian.net/browse/BOAS-315

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/create/key-dates/applications-create-key-dates.validators.js
+++ b/apps/web/src/server/applications/create/key-dates/applications-create-key-dates.validators.js
@@ -45,7 +45,7 @@ export const validateApplicationsCreateKeyDates = createValidator(
 			const { submissionInternalDay, submissionInternalMonth, submissionInternalYear } = req.body;
 			const submissionInternalDate = new Date(
 				submissionInternalYear,
-				submissionInternalMonth - 1,
+				Number.parseInt(submissionInternalMonth, 10) - 1,
 				submissionInternalDay
 			);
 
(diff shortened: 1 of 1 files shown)


apps/web/src/server/applications/create/key-dates/applications-create-key-dates.validators.js
```diff
@@ -45,7 +45,7 @@ export const validateApplicationsCreateKeyDates = createValidator(
 			const { submissionInternalDay, submissionInternalMonth, submissionInternalYear } = req.body;
 			const submissionInternalDate = new Date(
 				submissionInternalYear,
-				submissionInternalMonth,
+				submissionInternalMonth - 1,
```
mayya-bond-kc, line 48: It seems that earlier in the file we are using `Number.parseInt(submissionInternalMonth, 10)` to do the same thing? can we potentially standardise here?

DavidJWhiteDev: can do.  do we need to parseInt on the year and day too, to be consistent here?

mayya-bond-kc: Not sure, it doesn't seem like the day and year need to undergo the same treatment.

DavidJWhiteDev: ah, because I need to make a calculation on the month value (changing from 1-12 to 0-11 in this case!).  OK, I will just do it for month.

mayya-bond-kc (approved):

francescobrenci-AR (approved):

PR #538, opened 2022-09-16
Boas 354 filter data when resuming create application

Xfarya:
## Describe your changes
filter the GET request so only the required data is returned when  resuming a case - mainly refactoring

BOAS-354 https://pins-ds.atlassian.net/jira/software/projects/BOAS/boards/172?selectedIssue=BOAS-354

## Type of change 🧩

- [] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

- [x ] I have double checked this work does not include any hardcoded secrets or passwords

### DIFF
--- a/apps/api/src/server/applications/application/application.service.js
+++ b/apps/api/src/server/applications/application/application.service.js
@@ -132,7 +132,7 @@ const defaultInclusions = {
 
 /**
  *
- * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, locationDescription?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
+ * @param {{subSector?: boolean | object, sector?: boolean | object, caseEmail?: boolean | object, keyDates?: boolean | object, geographicalInformation?: boolean | object, locationDescription?: boolean | object, regions?: boolean | object, status?: boolean | object, applicants?: boolean | object, applicantsAddress?: boolean | object}} query
  * @returns {object}
  */
 const inclusionsUsingQuery = (query) => {
@@ -155,7 +155,7 @@ const inclusionsUsingQuery = (query) => {
 
 /**
  *
- * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
+ * @param {{subSector?: boolean | object, sector?: boolean | object, caseEmail?: boolean | object, keyDates?: boolean | object, geographicalInformation?: boolean | object, regions?: boolean | object, status?: boolean | object, applicants?: boolean | object, applicantsAddress?: boolean | object}} query
  * @returns {object}
  */
 const findModelsToInclude = (query) => {
@@ -172,8 +172,6 @@ const filterOutResponse = (parsedQuery, applicationDetailsFormatted) => {
 	const findTruthyValues = pickBy({ id: true, ...parsedQuery }, (value) => {
 		return value !== false;
 	});
-
-	// lodash.pick({test: {me: 'hello'}}, ['test.me'])
 
 	const findKey = Object.keys(findTruthyValues);
 
--- a/apps/api/src/server/applications/application/application.service.js
+++ b/apps/api/src/server/applications/application/application.service.js
@@ -130,6 +130,11 @@ const defaultInclusions = {
 	gridReference: true
 };
 
+/**
+ *
+ * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
+ * @returns {object}
+ */
 const inclusionsUsingQuery = (query) => {
 	return {
 		subSector: notFalseOrUndefined(query?.subSector) || notFalseOrUndefined(query?.sector),
@@ -142,15 +147,15 @@ const inclusionsUsingQuery = (query) => {
 		regions:
 			notFalseOrUndefined(query?.regions) || notFalseOrUndefined(query?.geographicalInformation),
 		caseStatus: query?.status,
-		serviceCustomer: notFalseOrUndefined(query.applicant),
-		serviceCustomerAddress: notFalseOrUndefined(query?.applicant?.address),
+		serviceCustomer: notFalseOrUndefined(query?.applicants),
+		serviceCustomerAddress: notFalseOrUndefined(query?.applicantsAddress),
 		gridReference: notFalseOrUndefined(query.geographicalInformation)
 	};
 };
 
 /**
  *
- * @param {{subSector?: any, sector?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicant?: any}} query
+ * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
  * @returns {object}
  */
 const findModelsToInclude = (query) => {
@@ -183,9 +188,7 @@ export const getCaseDetails = async (id, query) => {
 
 	const parsedQuery = !isEmpty(query) ? JSON.parse(query.query) : emptyQuery;
 	const modelsToInclude = findModelsToInclude(parsedQuery);
-
 	const caseDetails = await caseRepository.getById(id, modelsToInclude);
-
 	const applicationDetailsFormatted = mapApplicationDetails(caseDetails);
 
 	return typeof parsedQuery !== 'undefined'
--- a/apps/api/src/server/applications/application/application.service.js
+++ b/apps/api/src/server/applications/application/application.service.js
@@ -132,7 +132,7 @@ const defaultInclusions = {
 
 /**
  *
- * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
+ * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, locationDescription?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
  * @returns {object}
  */
 const inclusionsUsingQuery = (query) => {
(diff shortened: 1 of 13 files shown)


apps/api/src/server/applications/application/application.service.js
```diff
@@ -163,6 +169,9 @@ const filterOutResponse = (parsedQuery, applicationDetailsFormatted) => {
 	const findTruthyValues = pickBy({ id: true, ...parsedQuery }, (value) => {
 		return value !== false;
 	});
+
+	// lodash.pick({test: {me: 'hello'}}, ['test.me'])
+
```
mayya-bond-kc, line 174: Unnecessary comment

apps/api/src/server/applications/application/application.service.js
```diff
@@ -127,26 +127,32 @@ const defaultInclusions = {
 	gridReference: true
 };
 
+/**
+ *
+ * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, locationDescription?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
```
mayya-bond-kc, line 132: We can probably replace `any` with `boolean | object` for all of these keys`

apps/api/src/server/applications/application/application.service.js
```diff
@@ -127,26 +127,32 @@ const defaultInclusions = {
...
+		serviceCustomer: notFalseOrUndefined(query?.applicants),
+		serviceCustomerAddress: notFalseOrUndefined(query?.applicantsAddress),
 		gridReference: notFalseOrUndefined(query.geographicalInformation)
 	};
 };
 
 /**
  *
- * @param {{subSector?: any, sector?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicant?: any}} query
+ * @param {{subSector?: any, sector?: any, caseEmail?: any, keyDates?: any, geographicalInformation?: any, regions?: any, status?: any, applicants?: any, applicantsAddress?: any}} query
```
mayya-bond-kc, line 155: Same as for `inclusionUsingQuery` we can probably replace all the `any`s with `boolean | object`

apps/web/src/server/applications/create/applications-create.service.js
```diff
@@ -69,12 +69,14 @@ export const createApplicationDraft = async (payload, session) => {
...
+ * optional query parameters can filter the returned data
  *
+ * @param {string | null} query
  * @param {number} id
  * @returns {Promise<Application>}
  */
-export const getApplicationDraft = async (id) => {
-	return get(`applications/${id}`);
+export const getApplicationDraft = async (id, query = null) => {
+	return get(`applications/${id}${query ? `?query=${query}` : ''}`);
```
mayya-bond-kc, line 79: Do we need to default `query` to null? I would have thought that `query` defaults to `undefined` automatically?

apps/web/src/server/applications/create/case/applications-create-case.controller.js
```diff
@@ -44,7 +44,7 @@ import {
 export async function viewApplicationsCreateCaseName(req, response) {
 	const { applicationId } = response.locals || {};
 	const { title, description } = applicationId
-		? await getApplicationDraft(applicationId)
+		? await getApplicationDraft(applicationId, '{"name":true, "description":true}')
```
mayya-bond-kc, line 47: A suggestion around how these strings are generated to ensure the syntax would be to write `JSON.stringify({name: true, description: true})`

mayya-bond-kc:
Looks like some Web tests are failing

lutamKC:
so, every web test is failing.
The reason for that is that you're mocking the url 'applications/123' but now you're requesting the url 'applications/123?query=something'.

The nocks() method should have something like a wildcard to fix this:
https://github.com/nock/nock/tree/v9.2.3#specifying-path

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
```diff
@@ -1,7 +1,7 @@
 import { findAddressListByPostcode } from '@planning-inspectorate/address-lookup';
 import { updateApplicationDraft } from '../applications-create.service.js';
 import * as applicationsCreateApplicantService from './applications-create-applicant.service.js';
-import { getNthApplicant } from './applications-create-applicant.service.js';
```
lutamKC, line 4: are we still using getNthApplicant somewhere? if not, we can delete it

apps/web/src/server/applications/create/applicant/applications-create-applicant.service.js
```diff
@@ -83,3 +83,22 @@ export async function getNthApplicant(applicationId, applicantNumber) {
...
+/**
+ * Returns the applicant matching id of a draft application
+ *
+ * @param {number} applicationId
+ * @param {number} applicantId
+ * @param {null | string} query
+ * @returns {Promise<Applicant|null>}
+ */
+export async function getApplicantById(applicationId, applicantId, query = null) {
+	const { applicants } = await getApplicationDraft(applicationId, query);
```
lutamKC, line 96: can't we say that `query` here is always 'applicants:true'?

apps/web/src/server/applications/create/case/applications-create-case.controller.js
lutamKC: another strategy could be just passing the keys as an array, something like `['name', 'description', 'applicants']` and let the `getApplicationDraft` service do the job of creating the query string

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
```diff
@@ -355,9 +359,9 @@ export async function updateApplicationsCreateApplicantWebsite(
  * @type {import('@pins/express').RenderHandler<ApplicationsCreateApplicantTelephoneNumberProps, {}, {}, {}, {}>}
  */
 export async function viewApplicationsCreateApplicantTelephoneNumber(req, response) {
-	const { applicationId } = response.locals;
+	const { applicationId, applicantId } = response.locals;
 
-	const applicant = await getNthApplicant(applicationId, 0);
+	const applicant = await getApplicantById(applicationId, applicantId, '{"applicants": true}');
```
lutamKC, line 364: see below about the query always being `applicant` for this service

lutamKC (approved):

lutamKC:
the function 'getApplicationDraft' is being used elsewhere in the code (in some middleware functions and in the 'check-your-answers' page). Probably need to add the filter there as well

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
```diff
@@ -68,9 +68,9 @@ export async function updateApplicationsCreateApplicantTypes({ path, session, bo
  * @type {import('@pins/express').RenderHandler<ApplicationsCreateApplicantOrganisationNameProps, {}, {}, {}, {}>}
  */
 export async function viewApplicationsCreateApplicantOrganisationName(req, response) {
-	const { applicationId } = response.locals;
+	const { applicationId, applicantId } = response.locals;
 
-	const applicant = await getNthApplicant(applicationId, 0);
```
lutamKC, line 73: are we still using somewhere the getNthApplicant function? if not you can remove it

lutamKC (changes requested):

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
DavidJWhiteDev: it is different for the address call

apps/web/src/server/applications/create/applicant/applications-create-applicant.service.js
DavidJWhiteDev: it is different for the address call

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
Xfarya: @DavidJWhiteDev ?

apps/web/src/server/applications/create/case/applications-create-case.controller.js
Xfarya: @DavidJWhiteDev

apps/web/src/server/applications/create/applications-create.service.js
Xfarya: @DavidJWhiteDev - why does this equal null?

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
DavidJWhiteDev: yeah, I only left it in in case we needed it in the future!

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
lutamKC: oh, right.

apps/web/src/server/applications/create/case/applications-create-case.controller.js
DavidJWhiteDev: I like the idea of passing a proper array, and converting it.  I have seen that in the code elsewhere.  would look much cleaner!

apps/web/src/server/applications/create/applicant/applications-create-applicant.controller.js
lutamKC: nah, you can remove it

apps/web/src/server/applications/create/case/applications-create-case.controller.js
lutamKC: so if have an array of, say, `filters`, it could be something like:
```
const filterObject = {}
for(const filter of filters) {
filterObject = {...filterObject, [filter]: true}
}
const query = JSON.stringify(filterObject)
```

lutamKC (changes requested):
branch name doesn't seem correct. should be feature/boas-354-etc...

apps/web/src/server/applications/create/applications-create.service.js
DavidJWhiteDev: there are existing calls to this that only have 1 argument.  Causes a syntax error.  Rather than changing all existing calls to add "..., null)" it is better to keep the default argument setting in the fn I think

PR #591, opened 2022-11-02
feat(storage): created post endpoint to create blobStoreUrl (BOAS-504)

Xfarya:
## Describe your changes
created post endpoint to create blobStoreUrl 

BOAS-504 https://pins-ds.atlassian.net/browse/BOAS-504

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/document-storage/src/server/app/__tests__/document-storage.test.js
+++ b/apps/document-storage/src/server/app/__tests__/document-storage.test.js
@@ -54,4 +54,13 @@ test('returns error if any field missing (caseRef, documentName, caseType, GUID)
 	});
 });
 
-// const resp = await request.post('/document-location').send([])
+test('returns error if no object passed', async (t) => {
+	const resp = await request.post('/document-location').send([]);
+
+	t.is(resp.status, 400);
+	t.deepEqual(resp.body, {
+		errors: {
+			'': 'Please enter an object {} with caseRef, documentName, caseType and GUID'
+		}
+	});
+});
--- a/apps/document-storage/src/server/app/__tests__/document-storage.test.js
+++ b/apps/document-storage/src/server/app/__tests__/document-storage.test.js
@@ -1,0 +1,57 @@
+import test from 'ava';
+import supertest from 'supertest';
+import app from '../../app.js';
+
+const request = supertest(app);
+
+test('returns blobStoreUrl given document info (caseRef, documentName, caseType, GUID)', async (t) => {
+	const resp = await request.post('/document-location').send([
+		{
+			caseType: 'application',
+			caseReference: '1',
+			documentName: 'PINS1.pdf',
+			GUID: 'D987654321'
+		},
+		{
+			caseType: 'appeal',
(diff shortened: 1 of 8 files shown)


apps/document-storage/src/server/swagger.js
```diff
@@ -14,24 +14,33 @@ const document = {
...
+				}
+			}
+		]
+	},
 	components: {}
 };
 
 const outputFile = './src/server/swagger-output.json';
-const endpointsFiles = ['./src/server/app.js'];
+const endpointsFiles = ['./src/server/app/routes.js'];
```
mayya-bond-kc, line 44: Remember we spoke about doing this using the `*` wildcards? Could you please add that here?

apps/document-storage/src/server/app/__tests__/document-storage.test.js
```diff
@@ -0,0 +1,57 @@
...
+				'Please provide a valid caseType. caseType must be either "appeal" or "application"',
+			'[0].caseReference':
+				'Please provide a valid caseReference. caseReference is not the same as ID',
+			'[0].documentName': 'Please provide a valid documentName',
+			'[0].GUID': 'Please provide a valid GUID'
+		}
+	});
+});
+
+// const resp = await request.post('/document-location').send([])
```
mayya-bond-kc, line 57: This should be a third test scenario I believe?

apps/document-storage/src/server/swagger.js
Xfarya: yes, I think we tried several versions, and this was the one that worked in the end, right?

Xfarya: perhaps, however currently - `const endpointsFiles = ['./src/server/app/*.routes.js'] `is not returning anything, and neither is ` const endpointsFiles = ['./src/server/app/**/*.routes.js'] `
do you have any other suggestions?

apps/document-storage/src/server/app/blob-store/service.js
```diff
@@ -93,5 +90,32 @@ export async function uploadBlob(
...
+export async function documentsCreateUrl(documents) {
+	documents.map((document) => {
+		if (document.caseType.match('application')) {
+			document.blobStoreUrl = buildApplicationURL(
+				document.caseReference,
+				document.GUID,
+				document.documentName
+			);
+		}
+		return documents;
```
Xfarya, line 108: added twice to pass linting checks - is there another way around this?

mayya-bond-kc: What does the error say?

Xfarya: it says the .map needs a return statement. but then I need an additional return statement for the function

mayya-bond-kc: I think it would make more sense here to return `document` instead of `documents`. https://www.w3schools.com/jsref/jsref_map.asp

PR #597, opened 2022-11-07
fix(web/applications): fixed issue with slashed in URL builder helper

francescobrenci-AR:
## Describe your changes

fixed issue with slashed in URL builder NJK helper

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/views/applications/dashboard.njk
+++ b/apps/web/src/server/views/applications/dashboard.njk
@@ -32,7 +32,7 @@
 				<div class="pins-dashboard-box">
 					<h3 class="govuk-heading-l pins-dashboard-box--title">Create new case</h3>
 					<a class="govuk-button govuk-button--secondary pins-dashboard-box--btn"
-					   href="{{ 'case-create'|url }}">Create new case here</a>
+					   href="{{ 'base-url'|url }}/create-new-case">Create new case here</a>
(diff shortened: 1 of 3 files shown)


apps/web/src/server/views/applications/dashboard.njk
```diff
@@ -32,7 +32,7 @@
 				<div class="pins-dashboard-box">
 					<h3 class="govuk-heading-l pins-dashboard-box--title">Create new case</h3>
 					<a class="govuk-button govuk-button--secondary pins-dashboard-box--btn"
-					   href="{{ 'case-create'|url }}">Create new case here</a>
+					   href="{{ 'base-url'|url }}/create-new-case">Create new case here</a>
```
lutamKC, line 35: this looks like a fast fix. 
I'm sure this problem of the double / comes from a recent fix and probably affects case-view as well. might be better to investigate what happened

lutamKC (approved):

apps/web/src/server/views/applications/dashboard.njk
francescobrenci-AR: we do need away to build urls outside of those case and `base-url` helps with that.
For now I'm addressing only the double slash in that specific URL

PR #606, opened 2022-11-10
feat(web/applications): add blob storage upload (boas-553)

lutamKC:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/client/components/file-uploader/_html.js
+++ b/apps/web/src/client/components/file-uploader/_html.js
@@ -55,10 +55,14 @@ export const buildErrorListItem = (error) => {
  * @param {Element} uploadForm
  */
 export const buildProgressMessage = ({ show }, uploadForm) => {
-	// TODO: do this
 	const progressHook = uploadForm.querySelector('.progress-hook');
+	/** @type {HTMLButtonElement | null} */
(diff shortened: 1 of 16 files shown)


apps/web/src/client/components/file-uploader/_html.js
```diff
@@ -1,43 +1,51 @@
...
 
 /**
- * @param {File} uploadedFile
- * @returns {string}
+ * @param {FileWithRowId} uploadedFile
+ * @returns {HTMLElement}
  */
 export const buildRegularListItem = (uploadedFile) => {
-	const fileRowId = `file_row_${uploadedFile.lastModified}_${uploadedFile.size}`;
+	const li = document.createElement('li');
```
lutamKC, line 23: creating the element and then filling it up with the innerHtml is the best approach here because returning a Node rather than a string allows me to use `rows.append()` rather than `rows.innerHTML+=`

they both work, but innerHTML removes all the EventListeners set on its elements, which meanse that two loops are necessary: a first one to insert all the html elements and a second one to set eventListeners.
With .append(NODE) I can inserting new html elements without unsetting the eventlisteners.

francescobrenci-AR (approved):

francescobrenci-AR (approved):

PR #642, opened 2022-12-01
feat(document-check-function): adds check-new-document function (BOAS…

mayya-bond-kc:
…-536)

## Type of change 🧩

- [ ] Documentation update


apps/document-check-function/check-new-document/index.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @param {{bindingData: {uri: string}}} context
+ * @param {import('node:stream').Readable} myBlob
+ */
+export const checkMyBlob = async (context, myBlob) => {
+	const isInfected = await scanStream(myBlob);
+	const machineAction = mapIsInfectedToMachineAction(isInfected);
+	const { caseId, guid } = getBlobCaseIdAndGuid(context.bindingData.uri);
+
+	await sendDocumentStateAction(guid, caseId, machineAction);
```
chrisdalykinandcarta, line 31: As part of SEC-04, we need to 'quarantine' documents which have been deemed unsafe. I think this is open to interpretation, because to some extent these files are 'quarantined' when they're first uploaded (in the sense that only this function can access them). It might be worth thinking about moving the file to a separate container where nothing has READ access and this function only has WRITE access.

chrisdalykinandcarta: We're also going to need to raise some alarm here, for now it might just be worth adding in a console.warning so that at a minimum we can subscribe to the log group for warning messages.

apps/document-check-function/check-new-document/index.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+/**
+ * @param {string} blobUri
+ * @returns {{caseId: string, guid: string}}
+ */
+const getBlobCaseIdAndGuid = (blobUri) => {
+	const uriParts = blobUri.split('/');
+
+	return { caseId: uriParts[1], guid: uriParts[2] };
```
chrisdalykinandcarta, line 19: Might be worth checking the length of uriParts here and logging the URI with an error if something goes wrong

apps/document-check-function/check-new-document/__tests__/check-new-document.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+const fileStream = new ReadableStream();
+
+const patchStub = sinon.stub().returns({ json: sinon.stub() });
+
+test.before('set up mocks', () => {
+	sinon.stub(got, 'patch').callsFake(patchStub);
+});
+
+test.serial("sends 'passed' machine action to back office if passed AV check", async (t) => {
```
chrisdalykinandcarta, line 16: Really love this scope of testing, awesome job!

apps/document-check-function/check-new-document/clam-av-client.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @param {import('node:stream').Readable} blobStream
+ * @returns {Promise<boolean>}
+ */
+export const scanStream = async (blobStream) => {
+	const clamScan = await new NodeClam().init({
+		debugMode: true,
+		clamdscan: {
+			host: '0.0.0.0',
```
chrisdalykinandcarta, line 11: I guess this is supposed to come from some config?

mayya-bond-kc: Nice catch yes

apps/document-check-function/check-new-document/index.js
mayya-bond-kc: Good shout. Adding something now

apps/document-check-function/check-new-document/clam-av-client.js
```diff
@@ -0,0 +1,19 @@
+import NodeClam from 'clamscan';
+
+/**
+ * @param {import('node:stream').Readable} blobStream
+ * @returns {Promise<boolean>}
+ */
+export const scanStream = async (blobStream) => {
+	const clamScan = await new NodeClam().init({
```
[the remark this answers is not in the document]
mayya-bond-kc: Good point

apps/document-check-function/check-new-document/index.js
```diff
@@ -0,0 +1,87 @@
...
+export const checkMyBlob = async (context, myBlob) => {
+	const documentUri = context.bindingData.uri;
+	const { caseId, guid } = getBlobCaseIdAndGuid(documentUri);
+
+	const isInfected = await scanStream(myBlob);
+
+	if (isInfected) {
+		context.error('Document did not pass AV checks');
+		await deleteDocument(documentUri, context);
+	}
```
chrisdalykinandcarta, line 82: If 'isInfected' is true and this code runs, should we also be executing the following commands too?

apps/document-check-function/check-new-document/index.js
```diff
@@ -0,0 +1,87 @@
...
+ * Once we save the same info to the Blob Metadata, let's get it from there
+ *
+ * @param {string} blobUri
+ * @returns {{caseId: string, guid: string}}
+ */
+const getBlobCaseIdAndGuid = (blobUri) => {
+	const uriParts = blobUri.split('/');
+
+	if (uriParts.length !== 5 || uriParts[1] !== 'applications') {
+		throw new Error(`Unexpected blob URI ${blobUri}`);
```
chrisdalykinandcarta, line 29: For expected errors like this (and for example, where you throw an error if the state change has already been completed), are the errors being propagated to the top-level? We want to make sure that the functions only 'throw' errors for things that we can potentially retry, because we want to be able to retry for transient errors (e.g. if clamav container is down) but not retry for expected errors.

chrisdalykinandcarta (approved):

PR #651, opened 2022-12-06
Feature/boas 537 document status

Xfarya:
## Describe your changes
should:
not have the doc model migration
+4 tests:
-- the GUID does not exist in the database
-- we send an unexpected machineAction in the body (as in one that does not exist in the state machine)
-- we send a machineAction that is not allowed given the document's current state
-- we don't provide a machineAction at all
-renamed stubs
refactored Controller fx - moved to Service
changed variable names from 'get'
made variable names clearer
changed req type from any
changed swagger path
removed undefined as second status parameter

BOAS-537 https://pins-ds.atlassian.net/browse/BOAS-537

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/application.validators.js
+++ b/apps/api/src/server/applications/application/application.validators.js
@@ -6,6 +6,7 @@ import {
 	validationErrorHandlerMissing
 } from '../../middleware/error-handler.js';
 import * as caseRepository from '../../repositories/case.repository.js';
+import * as documentRepository from '../../repositories/document.repository.js';
 import * as folderRepository from '../../repositories/folder.repository.js';
 import * as regionRepository from '../../repositories/region.repository.js';
 import * as serviceCustomerRepository from '../../repositories/service-customer.repository.js';
@@ -76,6 +77,33 @@ const validateExistingMapZoomLevel = async (value) => {
 
 	if (mapZoomLevel == null) {
 		throw new Error('Unknown map zoom level');
+	}
+};
+
+/**
+ * @param {string} value
+ */
+const validateDocumentGUIDExistsInTable = async (value) => {
+	const documentGUID = await documentRepository.getByDocumentGUID(value);
+
+	if (documentGUID === null || typeof documentGUID === 'undefined') {
+		throw new Error('DocumentGUID must exist in database');
+	}
+};
+
+/**
+ * @param {string} value
+ * @param {import('express-validator').Meta} meta
+ */
+const validateDocumentGUIDBelongsToCase = async (value, { req }) => {
+	const documentGUID = await documentRepository.getByDocumentGUID(value);
+
+	const getCaseById = await folderRepository.getById(documentGUID?.folderId);
+
+	const caseId = getCaseById?.caseId;
+
+	if (Number.parseInt(req.params?.caseId, 10) !== caseId) {
+		throw new Error('GUID must belong to correct case');
 	}
 };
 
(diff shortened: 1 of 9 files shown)


apps/api/src/server/applications/application/application.validators.js
```diff
@@ -226,3 +253,17 @@ export const validateFolderIds = composeMiddleware(
 		.withMessage('Folder must belong to case'),
 	validationErrorHandler
 );
+
+export const validateDocumentGUID = composeMiddleware(
+	param('documentGUID')
+		.custom(validateDocumentGUIDExistsInTable)
+		.withMessage('DocumentGUID must exist in database')
+		.custom(validateDocumentGUIDBelongsToCase)
```
mayya-bond-kc, line 261: I don't think we need to have separate validators here. `validateDocumentGUIDBelongsToCase` will inherently check that the GUID exists in the DB. So I would remove `validateDocumentGUIDExistsInTable` altogether.

apps/api/src/server/repositories/document.repository.js
```diff
@@ -27,3 +27,25 @@ export const update = (documentGuid, documentDetails) => {
...
+ * @returns {import('@prisma/client').PrismaPromise<import('@pins/api').Schema.Document | null>}
+ */
+export const getByDocumentGUID = (documentGUID) => {
+	return databaseConnector.document.findUnique({
+		where: { guid: documentGUID }
+	});
+};
+
+/**
+ * @param {{guid: string, status: import('xstate').StateValue }} documentStatusUpdate
```
mayya-bond-kc, line 43: I am not sure it makes sense to sent a `StateValue` as the new status. A `string` makes more sense surely?

apps/api/src/server/repositories/folder.repository.js
```diff
@@ -21,7 +21,7 @@ export const getByCaseId = (caseId, parentFolderId = null) => {
 /**
  * Returns a single folder on a case
  *
- * @param {number} folderId
+ * @param {number | any} folderId
```
mayya-bond-kc, line 24: We should not allow `any` here

apps/api/src/server/applications/application/application.validators.js
Xfarya: yes I agree - however, for the tests, the validation error message is the same unless I separate these functions

apps/api/src/server/repositories/folder.repository.js
Xfarya: what would be the alternative - can I let 'undefined'?

apps/api/src/server/repositories/document.repository.js
Xfarya: agree - I changed it to string

apps/api/src/server/repositories/folder.repository.js
Xfarya: It gives the red squiggly line if I don't 🙈

mayya-bond-kc: If a TSCheck error appears, then we need to understand why and fix the source of the issue

apps/api/src/server/applications/application/application.validators.js
Xfarya: If we are looking to reduce the number of calls we make to the db, then we can get rid of it.
If that doesn't matter, then to keep it would provide more precise feedback?

I am happy to remove it - only because I feel there is a little unnecessary repetition of code.

apps/api/src/server/repositories/folder.repository.js
Xfarya: It says something along the lines of cannot use type 'number' on 'undefined'

apps/api/src/server/applications/application/application.validators.js
Xfarya: done

apps/api/src/server/repositories/folder.repository.js
Xfarya: i removed it - the tschecks are passing - but the squiggly lines are still there

francescobrenci-AR (approved):

PR #670, opened 2023-01-12
Feature/boas 694 download preview files

lutamKC:
## Describe your changes

Clicking on _download_ or _preview_ in the files list now opens an internal page that makes an URI with the blobstorage path and a short-term SAS token.
The SAS token is created with AD credentials.
The URI of the file needs the blobstoragePath and blobstorageContainer of the file which is returned by a new API.
Previewing the file will open it in the browser (when possible) in the same tab; dowloading it will just start the download.

Note:
The download/preview link for now is active whatever the status of document is (including virus-check-failed). This is to be changed in a different task.

https://pins-ds.atlassian.net/jira/software/projects/BOAS/boards/172?selectedIssue=BOAS-694

## Type of change 🧩

- [ ] Documentation update


apps/api/src/server/applications/application/documents/document.controller.js
```diff
@@ -82,3 +82,28 @@ export const updateDocuments = async ({ body }, response) => {
...
+ */
+export const getDocumentUri = async ({ params }, response) => {
+	let /** @type { import('apps/api/prisma/schema.js').Document |null} */ document = null;
+
+	try {
+		document = await documentRepository.getById(params.guid);
+
+		if (document === null || typeof document === 'undefined') {
+			throw new Error(`Unknown document guid ${params.guid}`);
+		}
```
mayya-bond-kc, line 100: This looks like it should live in a validator function in the `routes`?

DavidJWhiteDev: Yes I was initially going to have a validation fn, but it would have to make the same getByID call as in this fn, and I saw little value in calling the same fn twice separately

mayya-bond-kc: I wonder if we can do some operations in the validator and then save the data (such as the case obtained from the DB) and then send it over to the controller

lutamKC: this can be a bit difficult to handle, but on the antivirus-check branch there is a validator named "validateDocumentGUID", which I believe does exactly what is happening here. 
And yes, apparently that one uses the getById fn so maybe it could be used/optimized?

DavidJWhiteDev: after discussion with Mayya, agreed to keep code as is for now, but have a dev chat about a consistent way forward for similar stuff in the future

DavidJWhiteDev (approved):

PR #676, opened 2023-01-16
Feature/boas 690 publishing api

LetsMakeItTechnical:
## Type of change 🧩

- [ ] Documentation update


francescobrenci-AR:
it should be merged to this branch `feature/boas-676-preview-and-publish-page`

apps/api/src/server/applications/application/__tests__/publish-application.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+test('publish an application/case', async (t) => {
+	const caseId = 1;
+
+	const response = await request.patch(`/applications/${caseId}/publish`);
+
+	t.is(response.status, 200);
+
+	t.deepEqual(response.body, {
+		publishedDate: mapDateStringToUnixTimestamp(new Date(1_649_319_144_000).toISOString())
```
chrisdalykinandcarta, line 83: I don't think we're testing the functionality of the method here, we're essentially just asserting that it was called. Would it be more useful to test against what we expect the string to be?

apps/api/src/server/applications/application/__tests__/publish-application.test.js
```diff
@@ -0,0 +1,96 @@
...
+	sinon.stub(Prisma.PrismaClient.prototype, '$transaction').returns([
+		{
+			publishedAt: new Date(1_649_319_144_000)
+		}
+	]);
+
+	sinon.useFakeTimers({ now: 1_649_319_144_000 });
+});
+
+test('publish an application/case', async (t) => {
```
mayya-bond-kc, line 74: I would add a few more test cases here. For example:
- Does an error get thrown if an invalid caseId is provided?
- Does the publishedAt date get updated if re-publishing?

apps/api/src/server/applications/applications.routes.js
```diff
@@ -142,4 +143,33 @@ router.get(
...
+        }
+        #swagger.responses[200] = {
+            description: 'ID of application',
+            schema: { publishedDate }
+        }
+    */
+	validateApplicationId,
+	validateApplicantId,
+	validateCreateUpdateApplication,
+	trimUnexpectedRequestParameters,
```
mayya-bond-kc, line 171: Do we need these?

apps/api/src/server/repositories/case.repository.js
```diff
@@ -232,6 +234,7 @@ const updateApplicationSansRegionsRemoval = ({
 	return databaseConnector.case.update({
 		where: { id: caseId },
 		data: {
+			publishedAt: publishedAt || null,
```
mayya-bond-kc, line 237: As we are adding a new piece for data to associate with a case, I don't think there is a matching migration for it

mayya-bond-kc (changes requested):

apps/api/src/server/repositories/case.repository.js
DavidJWhiteDev: I was wondering if that was the case as well, Mayya

apps/api/src/server/repositories/case.repository.js
```diff
@@ -337,6 +340,38 @@ export const updateApplication = ({
...
+		where: { id: caseId },
+		data: {
+			publishedAt: new Date(),
+			modifiedAt: new Date()
+		}
+	});
+
+	transactions.push(application);
+
+	return databaseConnector.$transaction(transactions);
```
mayya-bond-kc, line 359: We don't actually need to use transactions here. Transactions are useful if we have multiple database operations that we would like to process, but here we are only doing a single `update` on `case`

LetsMakeItTechnical: ok will change this now

apps/api/src/server/applications/applications.routes.js
```diff
@@ -142,4 +143,25 @@ router.get(
...
+        #swagger.description = 'publish application'
+        #swagger.parameters['id'] = {
+            in: 'path',
+			description: 'Application ID',
+			required: true,
+			type: 'integer'
+		}
+        #swagger.responses[200] = {
+            description: 'response will have the date that the case was published',
+            schema: { publishedDate }
```
DavidJWhiteDev, line 160: this does not display any response info in the swagger api, I would suggest:

#swagger.responses[200] = {
            description: 'response will have the date that the case was published as a timestamp',
            schema: { "publishedDate": 1673873105 }
        }

lutamKC (approved):

mayya-bond-kc (approved):
LGTM

PR #688, opened 2023-01-23
fix(api/applications): fix seeded cases display error (BOAS-704)

DavidJWhiteDev:
## Describe your changes

These are fixes to the seeding script, and therefore only affect local, dev, and test environments, not Production.
1. This fixes seeding case not displaying correctly - both in draft, and for started cases, when trying to amend Application Info details.  
2. all seeded cases are assigned a caseReference eg EN010001.  This should not be the case for Draft cases, as cases are assigned a caseReference when they progress from Draft to Pre-application.  Code amended to set caseReference to NULL for Draft seeded cases.
3. case references have changed from a 4 digit number to include a leading ‘1' to separate these cases from existing Horizon case references. eg EN010001 should be EN0110001.  Seed code still generates 8 char ref codes, needs to change to 9 to include the '1’ after the SubSector code - ie EN01 +1. 

**Note that `npm run db:seed ` will need to be run after this has been delivered for these changes to take affect**

## BOAS-704 Seeded Draft Application cases cannot display
https://pins-ds.atlassian.net/browse/BOAS-704

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/prisma/seed.js
+++ b/apps/api/prisma/seed.js
@@ -38,14 +38,14 @@ function generateAppealReference() {
 	return `APP/Q9999/D/21/${number}`;
 }
 
-// Application reference should be in the format (subSector)(4 digit sequential_number) eg EN010001
+// Application reference should be in the format (subSector)(5 digit sequential_number with leading 1) eg EN0110001
 /**
  * @param {{abbreviation: string}} subSector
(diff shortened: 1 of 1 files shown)


mayya-bond-kc (approved):
Nice catch!

PR #689, opened 2023-01-23
build(web/applications): upgrade msal and express

lutamKC:
## Describe your changes

Upgrade express and MSAL + fix jest timeout problems due to `Throw New Error` usage

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/app/app.router.js
+++ b/apps/web/src/server/app/app.router.js
@@ -3,6 +3,7 @@ import { Router as createRouter } from 'express';
 import { installAuthMock } from '../../../testing/app/mocks/auth.js';
 import appealsRouter from '../appeals/appeals.router.js';
 import applicationsRouter from '../applications/applications.router.js';
+import asyncRoute from '../lib/async-route.js';
 import { handleHeathCheck, viewHomepage, viewUnauthenticatedError } from './app.controller.js';
 import { handleSignout } from './auth/auth.controller.js';
 import { assertIsAuthenticated } from './auth/auth.guards.js';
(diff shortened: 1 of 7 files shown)


apps/web/src/server/app/app.router.js
```diff
@@ -32,7 +33,7 @@ if (!config.authDisabled) {
 }
 
 router.route('/').get(viewHomepage);
-router.route('/auth/signout').get(handleSignout);
+router.route('/auth/signout').get(asyncRoute(handleSignout));
 router.route('/documents/:caseId/upload').post(postDocumentsUpload);
```
lutamKC, line 37: I didn't add the asyncRoute everywhere because in some case it was generating issues, but I'm definetely going to do that in a specific ticket for that.

francescobrenci-AR: ok - if that's not creating issues we can review it later

apps/web/src/server/applications/pages/create-new-case/case/__tests__/applications-create-case.test.js
```diff
@@ -217,17 +222,6 @@ describe('applications create', () => {
 						expect(element.innerHTML).not.toContain('checked');
 					});
 				});
-				describe('Not provided', () => {
```
francescobrenci-AR, line 220: don't you need to test this anymore?

lutamKC: no, this was created before the draft/non-draft guard was implemented. Now, there is a default beahviour (an error, of course) that is triggered when the ID of the case doesn't match with any draft case.
This test was making sure that, if you don't provide an ID to the "create-new-case/sector" page, you get an error, which is correct, but now the error you get is different and this test doesnt really make much sense anymore.

francescobrenci-AR (approved):

PR #703, opened 2023-01-27
feat(web): implement auth for internal api and added caching for acce…

LetsMakeItTechnical:
…ss token  (BOAS-469)

## Type of change 🧩

- [ ] Documentation update


apps/api/.env.development
```diff
@@ -31,3 +31,7 @@ SERVICE_BUS_SUBSCRIBER="back-office-api-subscriber"
 # - - - - - - - - - - - Document Storage API
 DOCUMENT_STORAGE_API_HOST="localhost"
 DOCUMENT_STORAGE_API_PORT=3001
+
```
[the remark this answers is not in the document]
LetsMakeItTechnical: yes it does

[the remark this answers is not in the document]
LetsMakeItTechnical: I need to sort out the infrastructure

apps/api/src/server/app.js
```diff
@@ -24,6 +25,10 @@ app.use(compression());
 app.use(morgan('combined'));
 app.use(helmet());
 
+if (process.env?.IS_CLIENT_CREDENTIAL_GRANT_FEATURE === 'true') {
```
mayya-bond-kc: For consistency, I would use the `config` object, which contains all the env vars

mayya-bond-kc: if we were to change this environment variable, we would need to reload the application to apply this middleware. It might make more sense to do this check inside the middleware itself, so that we would not need to restart the app

apps/api/src/server/middleware/auth-handler.js
```diff
@@ -0,0 +1,72 @@
...
+			accessToken = req.headers.authorization.split(' ')[1];
+		}
+
+		if (!accessToken) {
+			throw new Error(errorMessage);
+		}
+
+		logger.info('attempting to decode access token');
+
+		const decoded = jwt.decode(accessToken, { complete: true });
```
mayya-bond-kc, line 42: This is not a requirement, but a suggestion: I would move these single lines of code to their own functions, such as 
```
const getDecodedAccessToken = (accessToken) => {
    return jwt.decode(accessToken, { complete: true })
}
```
the reason I am suggesting this is because this function has many responsibilities and there is difficult to digest when reading it

apps/api/src/server/app.js
LetsMakeItTechnical: ok

apps/api/src/server/middleware/auth-handler.js
LetsMakeItTechnical: will do that makes sense

PR #718, opened 2023-02-07
feat(api/applications): caching for sector, region and zoomlevel (BOA…

Xfarya:
…S-468)

## Describe your changes
caching using node-cache, + tests

https://pins-ds.atlassian.net/jira/software/projects/BOAS/boards/172?selectedIssue=BOAS-468

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/region/__tests__/get-regions.test.js
+++ b/apps/api/src/server/applications/region/__tests__/get-regions.test.js
@@ -43,7 +43,7 @@ describe('Get regions', () => {
 	});
 });
 
-test('checks cache working for regions', async () => {
+test('tests if cached data for region call is returned without hitting db', async () => {
 	const cachedRegion = {
 		id: 1,
 		name: 'cached test',
--- a/apps/api/src/server/applications/region/__tests__/get-regions.test.js
+++ b/apps/api/src/server/applications/region/__tests__/get-regions.test.js
@@ -1,5 +1,7 @@
 import supertest from 'supertest';
 import { app } from '../../../app.js';
+import { nodeCache, setCache } from '../../../utils/cache-data.js';
+
 const { databaseConnector } = await import('../../../utils/database-connector.js');
 
 const request = supertest(app);
@@ -30,4 +32,29 @@ describe('Get regions', () => {
 			}
 		]);
 	});
+});
+
+test('checks cache working for regions', async () => {
+	nodeCache.flushAll();
+
+	const cachedRegion = {
+		id: 1,
+		name: 'cached test',
+		displayNameEn: 'cached test name en',
+		displayNameCy: 'cached test name cy'
+	};
+
+	setCache('regions', [cachedRegion]);
+
+	const resp = await request.get('/applications/region');
+
+	expect(resp.status).toEqual(200);
+	expect(resp.body).toEqual([
+		{
+			id: cachedRegion.id,
+			name: cachedRegion.name,
+			displayNameEn: cachedRegion.displayNameEn,
+			displayNameCy: cachedRegion.displayNameCy
+		}
+	]);
 });
--- a/apps/api/src/server/applications/region/__tests__/get-regions.test.js
+++ b/apps/api/src/server/applications/region/__tests__/get-regions.test.js
@@ -1,3 +1,4 @@
+import { jest } from '@jest/globals';
 import supertest from 'supertest';
 import { app } from '../../../app.js';
 import { nodeCache, setCache } from '../../../utils/cache-data.js';
@@ -14,6 +15,14 @@ const region = {
 };
 
 describe('Get regions', () => {
+	beforeEach(() => {
+		nodeCache.flushAll();
+	});
+
+	afterEach(() => {
+		jest.resetAllMocks();
+	});
+
 	test('gets all regions', async () => {
 		// GIVEN
 		databaseConnector.region.findMany.mockResolvedValue([region]);
@@ -35,8 +44,6 @@ describe('Get regions', () => {
 });
 
 test('checks cache working for regions', async () => {
-	nodeCache.flushAll();
-
 	const cachedRegion = {
 		id: 1,
 		name: 'cached test',
@@ -48,6 +55,7 @@ test('checks cache working for regions', async () => {
 
 	const resp = await request.get('/applications/region');
 
+	expect(databaseConnector.region.findMany).not.toHaveBeenCalled();
 	expect(resp.status).toEqual(200);
 	expect(resp.body).toEqual([
 		{
--- a/apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
+++ b/apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
@@ -82,7 +82,6 @@ describe('Get sectors', () => {
 			.get('/applications/sector')
 			.query({ sectorName: unknownSectorName });
 
-
 		// THEN
 		expect(response.status).toEqual(400);
 		expect(response.body).toEqual({
@@ -119,6 +118,8 @@ test('test if sector cache is working', async () => {
 	]);
 });
 
+const sectorName = 'Test1';
+
 test('test if cache is working with specific string', async () => {
 	nodeCache.flushAll();
 
@@ -132,7 +133,9 @@ test('test if cache is working with specific string', async () => {
 
 	setCache('Test1', [cacheSector]);
 
-	const response = await request.get('/applications/sector').query({ sectorName: 'Test1' });
+	databaseConnector.sector.findUnique.mockResolvedValue([sector]);
+
+	const response = await request.get('/applications/sector').query({ sectorName });
 
 	expect(response.status).toEqual(200);
 	expect(response.body).toEqual([
@@ -158,7 +161,9 @@ test('test if cache is working with specific string', async () => {
 
 	setCache('Test1', [cacheSector]);
 
-	const response = await request.get('/applications/sector').query({ sectorName: 'Test1' });
+	databaseConnector.sector.findUnique.mockResolvedValue([sector]);
+
+	const response = await request.get('/applications/sector').query({ sectorName });
 
 	expect(response.status).toEqual(200);
 	expect(response.body).toEqual([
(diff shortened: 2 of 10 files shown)


apps/api/src/server/applications/sector/sector.controller.js
```diff
@@ -21,12 +22,18 @@ const mapSectors = (sectors) => {
...
 
-	// sort ascending order of subsector abbreviation, BC, EN, ... WA, WS, WW
-	sectors.sort((a, b) => (a.abbreviation > b.abbreviation ? 1 : -1));
+	if (!sectors) {
+		sectors = isEmpty(request.query)
+			? await sectorRepository.getAll()
+			: await subSectorRepository.getBySector({ name: request.query.sectorName?.toString() });
+
+		// sort ascending order of subsector abbreviation, BC, EN, ... WA, WS, WW
+		sectors.sort((a, b) => (a.abbreviation > b.abbreviation ? 1 : -1));
```
mayya-bond-kc, line 33: Something that we should be doing is sorting in the database, so let's update the call to the database such that we receive a sorted list back

apps/api/src/server/applications/sector/sector.controller.js
```diff
@@ -21,12 +22,18 @@ const mapSectors = (sectors) => {
  * @type {import('express').RequestHandler}
  */
 export const getSectors = async (request, response) => {
-	const sectors = isEmpty(request.query)
-		? await sectorRepository.getAll()
-		: await subSectorRepository.getBySector({ name: request.query.sectorName?.toString() });
+	let sectors = await getCache(request.query.sectorName?.toString() || '');
```
mayya-bond-kc, line 25: Looks like a bit of repeating code with `request.query.sectorName?.toString()`

apps/api/src/server/applications/sector/sector.controller.js
```diff
@@ -21,12 +22,18 @@ const mapSectors = (sectors) => {
...
-	sectors.sort((a, b) => (a.abbreviation > b.abbreviation ? 1 : -1));
+	if (!sectors) {
+		sectors = isEmpty(request.query)
+			? await sectorRepository.getAll()
+			: await subSectorRepository.getBySector({ name: request.query.sectorName?.toString() });
+
+		// sort ascending order of subsector abbreviation, BC, EN, ... WA, WS, WW
+		sectors.sort((a, b) => (a.abbreviation > b.abbreviation ? 1 : -1));
+
+		setCache(request.query.sectorName?.toString() || '', sectors);
```
mayya-bond-kc, line 35: This can be its own whole function and is responsible for getting the sectors from the database. In theory each function should only be a few lines long

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
```diff
@@ -89,3 +91,87 @@ describe('Get sectors', () => {
...
+
+	const cacheSector = {
+		id: 3,
+		name: 'cache test name',
+		abbreviation: 'CC',
+		displayNameEn: 'cache test name en',
+		displayNameCy: 'cache test name cy'
+	};
+
+	setCache('', [cacheSector]);
```
mayya-bond-kc, line 106: How come we are setting cache to an empty string here?

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
```diff
@@ -89,3 +91,87 @@ describe('Get sectors', () => {
...
+			displayNameEn: cacheSector.displayNameEn,
+			displayNameCy: cacheSector.displayNameCy
+		}
+	]);
+});
+
+const sectorName = 'Test1';
+
+test('test if cache is working with specific string', async () => {
+	nodeCache.flushAll();
```
mayya-bond-kc, line 124: If we have repeating things like `nodeCache.flushAll()` when preparing tests, I would recommend using a `beforeEach` to avoid repeating code

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
```diff
@@ -89,3 +91,87 @@ describe('Get sectors', () => {
...
+
+	expect(response.status).toEqual(200);
+	expect(response.body).toEqual([
+		{
+			name: cacheSector.name,
+			abbreviation: cacheSector.abbreviation,
+			displayNameEn: cacheSector.displayNameEn,
+			displayNameCy: cacheSector.displayNameCy
+		}
+	]);
```
mayya-bond-kc, line 176: I would say that here to double check our word, we should check that the `databaseConnector.sector.findUnique` function has definitely not been called

apps/api/src/server/utils/cache-data.js
```diff
@@ -0,0 +1,14 @@
+import NodeCache from 'node-cache';
+
+export const nodeCache = new NodeCache();
+
+export const setCache = (
+	/** @type {NodeCache.Key} */ key,
+	/** @type {Record<string,any>[]} */ value
+) => {
+	nodeCache.set(key, value, 3_600_000);
```
mayya-bond-kc, line 9: Can we put a comment here on how long the cache will be valid for given 3_600_000?

Xfarya: we set it to 1h for now - how long do you think we should store it?

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
Xfarya: I can't remember - @LetsMakeItTechnical - what was the reason?

apps/api/src/server/utils/cache-data.js
mayya-bond-kc: 1 hour sounds fine. I wasn't sure if this was in seconds/minutes etc, so a comment to say that on here would be great

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
Xfarya: 'double check our word' - what do you mean? It is returning the cacheSector object in this test - if you mean to remove the findUnique fx, if I do that, the test does not pass.

mayya-bond-kc: lol i have no idea what that sentence was mean to be :D 

Yes, i can see that the returned data is as we would expect. How come the tests don't pass?

mayya-bond-kc: lol i have no idea what that sentence was mean to be :D 

Yes, i can see that the returned data is as we would expect. How come the tests don't pass?

apps/api/src/server/applications/sector/sector.controller.js
Xfarya: I don't understand. This (setCache) is already a function stored in the utils folder?

mayya-bond-kc: Yes, sorry I mean everything inside the `if` clause

apps/api/src/server/applications/region/__tests__/get-regions.test.js
```diff
@@ -31,3 +33,28 @@ describe('Get regions', () => {
...
+
+	expect(resp.status).toEqual(200);
+	expect(resp.body).toEqual([
+		{
+			id: cachedRegion.id,
+			name: cachedRegion.name,
+			displayNameEn: cachedRegion.displayNameEn,
+			displayNameCy: cachedRegion.displayNameCy
+		}
+	]);
```
mayya-bond-kc, line 59: We can do the same thing here and ensure that `database.region.findMany` has not been called

apps/api/src/server/applications/sector/sector.controller.js
Xfarya: I added the sort fx to a file in the utils folder - called it 'sortAll' - then I wrap that fx around 'getAll' in sectorRepository, and 'getBySector' in subSectorRepository -- is this correct?

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
Xfarya: the tests are passing with the above - it returns 400 if I remove `databaseConnector.sector.findUnique.mockResolvedValue([sector]);` as I guess this is the call to the db

apps/api/src/server/applications/sector/sector.controller.js
mayya-bond-kc: So what I mean is that there are generally two spaces we can sort things: in the memory of the application itself or in the database. Ideally, we want to do all sorting in the database, so the way we can do that is by send to prisma that we would like the items sorted. So moving the code to a different space is not going to help us here

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
mayya-bond-kc: Yeah looks like a bug to me

Xfarya: so, do you mean to find a way to resolve this that does not use (i.e. remove) this line: `databaseConnector.sector.findUnique.mockResolvedValue([sector]);` ?

apps/api/src/server/applications/region/__tests__/get-regions.test.js
Xfarya: I am not sure what the 'same thing' is?

apps/api/src/server/applications/sector/__tests__/get-sectors.test.js
mayya-bond-kc: Essentially I think yes. Basically we want to test that we don't make any calls to the database at all

apps/api/src/server/applications/region/__tests__/get-regions.test.js
mayya-bond-kc: Sorry, I meant that we can test here same as we do in the "get regions" endpoint that we do not make any calls to a database

apps/api/src/server/applications/zoom-level/__tests__/get-zoom-map.test.js
```diff
@@ -33,3 +44,29 @@ describe('Get zoom map', () => {
 		]);
 	});
 });
+
+test('tests if cache is working', async () => {
```
mayya-bond-kc, line 48: I think these can be a little more descriptive. "Tests if cache is working" does not say exactly what the cache need to be doing

mayya-bond-kc (approved):
LGTM

mayya-bond-kc (approved):

PR #742, opened 2023-02-23
feat(api): Create and Update document metadata

LetsMakeItTechnical:
## Describe your changes

**Description**

This pull request adds a function storeDocumentMetadata that creates or updates a document metadata record in the database. The function handles an HTTP request and responds with the stored metadata upon successful storage.

**Changes Made**

Created a function named storeDocumentMetadata
Added @async, @function, @name, and @type annotations to describe the function
Added parameters for HTTP request (request) and response (response)
Added @throws annotation to describe the error that can be thrown by the function
Defined the request handler type as import('express').RequestHandler<{id: string; guid:string;}, ?, ?, any>
Extracted id and guid from the request parameters
Extracted the document metadata from the request body
Fetched the document from the database using the fetchDocumentByGuidAndCaseId function
Upserted the document metadata with the documentGuid property set to the document's GUID
Sent the stored metadata as the response body with a status code of 200

**Testing**
Tested the function with different input combinations to verify successful storage and error handling.

## Type of change 🧩

- [x] Documentation update


DavidJWhiteDev (changes requested):
DocumentMetaData table - some of these meta data / properties are already in the Document table - eg fileSize, FileType (mime type string), date Created etc.  Would it not be better to just extend the Document table to store the additional properties, rather than make a new table as well as the Document table?

LetsMakeItTechnical:
> DocumentMetaData table - some of these meta data / properties are already in the Document table - eg fileSize, FileType (mime type string), date Created etc. Would it not be better to just extend the Document table to store the additional properties, rather than make a new table as well as the Document table?

Storing document metadata in a separate table has several advantages, particularly when it comes to version control. By separating the metadata from the document itself, we can avoid the need to duplicate the document every time we make a change to its metadata.

For example, if we were to store the metadata directly in the document table and wanted to update a metadata field, we would have to create a new row for the document with the updated metadata, even if the document itself hasn't changed. This could lead to a large amount of unnecessary data duplication and make it difficult to track changes to the document metadata over time.

By storing the metadata in a separate table, we can update the metadata without having to create a new document row, which helps to reduce data duplication and make it easier to track changes over time. Additionally, separating the metadata from the document can make it easier to manage permissions, since we can set different permissions for the document and its metadata.

Overall, using a separate table for document metadata can provide better version control, reduce data duplication, and make it easier to manage permissions for both the document and its metadata.

LetsMakeItTechnical:
> DocumentMetaData table - some of these meta data / properties are already in the Document table - eg fileSize, FileType (mime type string), date Created etc. Would it not be better to just extend the Document table to store the additional properties, rather than make a new table as well as the Document table?

We need a new table for metadata because

apps/api/prisma/schema.prisma
```diff
@@ -306,18 +306,63 @@ model Folder {
...
+  securityClassification String?
+  sourceSystem           String
+  origin                 String
+
+  owner  String?
+  author String?
+
+  representative String
+  description    String
+  stage          Int?
```
mayya-bond-kc, line 365: Will this refer to the stage of the case during upload? If so, I would say that this could be a little more descriptive. Otherwise it can be confused with Document Status

apps/api/prisma/schema.prisma
```diff
@@ -306,18 +306,63 @@ model Folder {
...
-  redacted             Boolean  @default(false)
-  isDeleted            Boolean  @default(false)
-  createdAt            DateTime @default(now())
-  fileSize             Int      @default(0)
+  status               String             @default("awaiting_upload")
+  redacted             Boolean            @default(false)
+  isDeleted            Boolean            @default(false)
+  createdAt            DateTime           @default(now())
+  fileSize             Int                @default(0)
   fileType             String?
```
mayya-bond-kc, line 320: We will need to remove these columns, because they are redundant to the ones in the Metadata table

apps/api/src/server/applications/application/documents/document.controller.js
```diff
@@ -97,3 +100,46 @@ export const deleteDocumentSoftly = async ({ params: { id: caseId, guid } }, res
...
+
+	const documentMetadata = await databaseConnector.documentMetadata.create({
+		...documentMetadataBody,
+		documentGuid: document.guid
+	});
+
+	response.status(201).send(documentMetadata);
+};
+
+// @TODO
```
mayya-bond-kc, line 129: Left over from dev work?

apps/api/src/server/applications/application/documents/document.routes.js
```diff
@@ -17,6 +19,58 @@ import {
...
+            in: 'path',
+			description: 'guid of the required document here',
+			required: true,
+			type: 'string'
+		}
+		#swagger.responses[201] = {
+			description: 'The metadata has been successfully stored.',
+			schema: { $ref: '#/definitions/storeMetadataResponse' }
+		}
+    */
```
mayya-bond-kc, line 70: Let's add a validation here for the data being sent over to the API and ensure only the expected parts arrive to the controller

apps/api/src/server/applications/application/documents/document.controller.js
LetsMakeItTechnical: getting metdata will come after

apps/api/prisma/schema.prisma
```diff
@@ -306,18 +306,63 @@ model Folder {
...
   @@unique([name, folderId])
 }
+
+model DocumentMetadata {
+  id Int @id @unique @default(autoincrement())
+
+  Document     Document? @relation(fields: [documentGuid], references: [guid])
+  documentGuid String
+
+  caseRef String?
```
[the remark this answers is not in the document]
LetsMakeItTechnical: makes sense

apps/api/prisma/schema.prisma
LetsMakeItTechnical: I need to look more into it and find out that why this is a draft pr so we can all discuss this

apps/api/prisma/schema.prisma
LetsMakeItTechnical: yeah I agree will need to do that slowly don't wan't to break anything

apps/api/prisma/schema.prisma
DavidJWhiteDev: Stage is a value that is applied to the document when it is uploaded to the system.  It marks what stage in the case this document applied to.  It will not change as the case progresses.

apps/api/prisma/schema.prisma
```diff
@@ -306,18 +306,63 @@ model Folder {
...
-  isDeleted            Boolean  @default(false)
-  createdAt            DateTime @default(now())
-  fileSize             Int      @default(0)
+  status               String             @default("awaiting_upload")
+  redacted             Boolean            @default(false)
+  isDeleted            Boolean            @default(false)
+  createdAt            DateTime           @default(now())
+  fileSize             Int                @default(0)
   fileType             String?
+  DocumentMetadata     DocumentMetadata[]
```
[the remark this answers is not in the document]
francescobrenci-AR: I don't believe there will be "common" metadata that are separated from any version - each version will have it's full set of metadata but will all be inherited from the previous version. the latest version will be the working version and you can edit it's metadata which will be propagated to the next version but can't edit previous versions metadata

mayya-bond-kc (dismissed):

apps/api/prisma/schema.prisma
```diff
@@ -316,8 +316,45 @@ model Document {
   redacted             Boolean  @default(false)
   isDeleted            Boolean  @default(false)
   createdAt            DateTime @default(now())
-  fileSize             Int      @default(0)
+  fileSize             Int?     @default(0)
```
PINSDigitalServices, line 319: if we are making fileSize nullable, do we need the default value now?  (I only added the default value 0 in my original version of the Document table, to avoid having a nullable field)

apps/api/prisma/schema.prisma
```diff
@@ -319,5 +319,42 @@ model Document {
...
+  documentMetadata DocumentMetadata? @relation(fields: [metadataId], references: [id])
+  metadataId       Int?              @unique
+
   @@unique([name, folderId])
 }
+
+model DocumentMetadata {
+  id                     Int       @id @unique @default(autoincrement())
+  version                Int
+  dateCreated            DateTime
```
DavidJWhiteDev, line 331: query for all devs really: do we want to be consistent on naming for things like date fields?  elsewhere we have used names like createdAt to hold the date created.  (Though tbh I think dateCreated is clearer!)

LetsMakeItTechnical: While 'dateCreated' may be clearer, I agree that we should prioritize consistency, so using 'createdAt' makes sense for the time being.

apps/api/prisma/schema.prisma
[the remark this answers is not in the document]
LetsMakeItTechnical: done

DavidJWhiteDev (approved):

DavidJWhiteDev (approved):
lgtm!

PR #827, opened 2023-03-31
Feature/asb 1198 api rel reps get by

DavidBeale:
## Describe your changes

Endpoint to get a representation of a case by id

## Type of change 🧩

- [ ] Documentation update


apps/api/src/server/applications/application/application.validators.js
```diff
@@ -55,6 +56,20 @@ const validateExistingApplication = async (value) => {
...
+ * @throws {Error}
+ * @returns {Promise<void>}
+ */
+const validateExistingRepresentation = async (value) => {
+	const representation = await representationRepository.getById(value);
+
+	if (representation === null || typeof representation === 'undefined') {
+		throw new Error('Unknown representation');
+	}
+};
```
mayya-bond-kc, line 71: So we have been moving away from making DB calls in the validators that we use before the api call reaches the controller. As such, we are validating the inputs at this stage to check that they exist and are in the right format. Checking if a representation is in the DB would then happen at the controller level. This is essentially because we came to an agreement that we would like to not make any duplicate API calls and with this approach, we can pass around responses from databases within the controller more easily.

apps/api/src/server/applications/applications.routes.js
```diff
@@ -248,11 +250,49 @@ router.get(
 		}
     */
 	validateApplicationId,
-	validateGetApplicationQuery,
 	validateGetRepresentationsQuery,
 	asyncHandler(getApplicationRepresentations)
 );
 
+router.get(
+	'/:id/representations/:repId',
```
mayya-bond-kc, line 258: I am going to guess that there are going to be more endpoints starting with `/:id/representations` and so I would suggest that we create a folder just for these endpoints and put the controller/endpoints there, following the example of the `/case-team` and other groups of endpoints at the start of this file.

apps/api/src/server/applications/application/application.service.js
```diff
@@ -222,6 +222,16 @@ export const getCaseRepresentations = async (caseId, pagination, filterAndSort)
 	return representationsRepository.getByCaseId(caseId, pagination, filterAndSort);
 };
 
+/**
+ *
+ * @param {number} caseId
+ * @param {number} repId
+ * @returns {Promise<any>}
```
mayya-bond-kc, line 229: If possible, could we avoid typing responses with `any`?

apps/api/src/server/applications/application/__tests__/get-application-representation.test.js
```diff
@@ -0,0 +1,26 @@
...
+		id: 1,
+		reference: 'BC0110001-2',
+		status: 'VALID',
+		redacted: true,
+		received: '2023-03-14T14:28:25.704Z'
+	}
+];
+
+describe('Get Application Representation', () => {
+	it('gets rep for a case by id', async () => {
```
mayya-bond-kc, line 18: Could we add some more tests cases here, such as if the case does not exist in the db?

apps/api/src/server/repositories/representation.repository.js
```diff
@@ -53,6 +53,92 @@ export const getByCaseId = async (caseId, { page, pageSize }, { searchTerm, filt
...
+			id,
+			...caseFilter
+		}
+	});
+
+	const representation = representations[0];
+
+	if (representation === null || typeof representation === 'undefined') {
+		throw new Error('Unknown representation on application');
+	}
```
mayya-bond-kc, line 137: I would say we need to remove this logic out of this function. A few reasons for this:
1. The validator defined in `application.validators.js` will never be invoked
2. For different endpoints, we might want to handle the missing representations differently (some might want to throw and error, some might want to swallow the error). 
3. We generally do not perform these checks within any other repository and it is therefore a little confusing to have differing behaviour for this repository.

mayya-bond-kc (changes requested):

apps/api/src/server/applications/application/__tests__/get-application-representation.test.js
jonathangoulding: addressed

apps/api/src/server/applications/application/application.service.js
jonathangoulding: addressed

apps/api/src/server/applications/applications.routes.js
jonathangoulding: i have moved this file into the applications folder.

jonathangoulding: addressed

apps/api/src/server/applications/application/application.validators.js
jonathangoulding: This is following the patterns in place. Similar to -all the other functions. 

Could you provide an example where you have moved away from this pattern and if all the validators have been refactored to the 'one api call only' approach. 

Would we need to change this single instance over or all of the functions in the file ? 

Feels like it should be addressed as one tech debt ticket.

apps/api/src/server/repositories/representation.repository.js
jonathangoulding: So the repository should returned the raw result ? Which could be undefined / null ? 

And then each calling function (such as getCaseRepresentation) would need to handle the response.

apps/api/src/server/applications/application/application.validators.js
mayya-bond-kc: I understand the confusion and I would have done the same in your position. An example I can provide is the endpoint in `apps/api/src/server/applications/application/documents/document.routes.js` on line 134 (`getDocumentProperties`). Within the `getDocumentProperties` controller we check if the document exists in the DB and throw an error if it does not. 

We are not entirely in a position to spend a lot of time addressing tech debt at this point, though it would be a discussion we could have with @francescobrenci-AR.

In essence, I agree that it would be idea to address instances of this across the codebase, but I am not certain we are in a position to do so.

apps/api/src/server/repositories/representation.repository.js
mayya-bond-kc: That is what I would lean towards yes

mayya-bond-kc (approved):

apps/api/src/server/repositories/representation.repository.js
jonathangoulding: i will address later today.

apps/api/src/server/repositories/representation.repository.js
```diff
@@ -53,6 +53,86 @@ export const getByCaseId = async (caseId, { page, pageSize }, { searchTerm, filt
...
+
+	if (caseId) {
+		caseFilter = {
+			case: {
+				id: caseId
+			}
+		};
+	}
+
+	const representations = await databaseConnector.representation.findMany({
```
mayya-bond-kc, line 73: Just an FYI, if you are looking to find only the first record in the DB for a certain table, then you can use the `findFirst` function instead of `findMany()[0]`

mayya-bond-kc (approved):
LGTM

PR #843, opened 2023-04-19
fix(api/applications): error setting no document type (BOAS-830)

DavidJWhiteDev:
## Describe your changes

Fix error when trying to set No Document Type on a document.

This fix changes the DB structure of DocumentVersion to allow documentType to be nullable.  This should NOT be a breaking change.  DB migrate will have to be performed to update local dev etc.

The change also now changes Web as well as API.

## BOAS-830 No Document Type Error (BOAS-757)
https://pins-ds.atlassian.net/browse/BOAS-830

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/prisma/migrations/20230420095018_documentversion_documenttype_nullable/migration.sql
+++ b/apps/api/prisma/migrations/20230420095018_documentversion_documenttype_nullable/migration.sql
@@ -1,0 +1,20 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[DocumentVersion] DROP CONSTRAINT [DocumentVersion_documentType_df];
+ALTER TABLE [dbo].[DocumentVersion] ALTER COLUMN [documentType] NVARCHAR(1000) NULL;
+
+COMMIT TRAN;
+
+END TRY
+BEGIN CATCH
+
+IF @@TRANCOUNT > 0
+BEGIN
+    ROLLBACK TRAN;
+END;
+THROW
+
+END CATCH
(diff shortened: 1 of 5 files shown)


mayya-bond-kc (commented):
I don't quire understand, so we would allow a blank document name?

mayya-bond-kc:
> > I don't quire understand, so we would allow a blank document name?
> 
> @mayya-bond-kc This change is only to allow Document Type to be optional: users can choose "Rule 8 letter", "Exam library" or "No document type" - which will put a blank string in the DB.

Aha, ok. I understand that we can have documents without a type, however storing an empty string seems to be a little wasteful in terms of space as opposed to storing nothing at all. Let me know if I am not getting something

DavidJWhiteDev:
> 

@mayya-bond-kc the DB field is set up to not allow null, and has  default value of "" in the prisma schema.  The front end passes "" in, so this fix is just to handle that in the validation

mayya-bond-kc:
I would then say that we can lessen this restriction in the database. No need to store empty strings then

DavidJWhiteDev:
> I would then say that we can lessen this restriction in the database. No need to store empty strings then

@mayya-bond-kc yeah that makes sense.  Though I think that should be a separate ticket as it is beyond the scope of this bug fix?

francescobrenci-AR:
> > I would then say that we can lessen this restriction in the database. No need to store empty strings then
> 
> @mayya-bond-kc yeah that makes sense. Though I think that should be a separate ticket as it is beyond the scope of this bug fix?

I reckon we can use this bug to make this improvement instead of opening a new ticket - you can add a comment on it to describe the changes

PR #876, opened 2023-04-28
fix(web): use bearer token instead of sas token for downloads

chrisdalykinandcarta:
## Describe your changes

Instead of creating a SAS token, use the bearer token from the user identity to download the blob content. We need to do this through the SDK rather than the public URL since the public URL only supports SAS tokens in query strings.

This was an issue because SAS tokens have to be scoped to a storage account (not just a container), but the AD groups which CAT team belong to only have RBAC roles assigned at the container level. More detail in the comments on JIRA.

BOAS-834

## Type of change 🧩

- [ ] Documentation update


lutamKC (approved):

lutamKC:
lgtm!

beejjacobs (approved):

PR #888, opened 2023-05-04
feat(api): add pagination for /appeals route

shepda:
## Describe your changes

Adds pagination for the `/appeals` route as query params.

For example `http://localhost:3000/appeals?pageNumber=1&pageSize=30`.

If no pagination params are given then the response is limited to the first 30 appeals.

If pagination params are given then:

- Both `pageNumber` and `pageSize` must be given
- Both `pageNumber` and `pageSize` must be numeric
- Both `pageNumber` and `pageSize` must be greater than 0

Also added tests for success and error states and updated swagger docs.

https://pins-ds.atlassian.net/browse/BOAT-41

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/appeals/appeals/appeals.validators.js
+++ b/apps/api/src/server/appeals/appeals/appeals.validators.js
@@ -1,10 +1,51 @@
 import { composeMiddleware } from '@pins/express';
-import { param } from 'express-validator';
+import { param, query } from 'express-validator';
 import { validationErrorHandler } from '../../middleware/error-handler.js';
+
+/**
+ * @param {string} value
+ * @returns {boolean}
+ */
+const hasValue = (value) => !!value;
+
+/**
+ * @param {string} value
+ * @returns {boolean}
+ */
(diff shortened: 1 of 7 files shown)


apps/api/src/server/appeals/appeals/appeals.validators.js
```diff
@@ -1,10 +1,51 @@
 import { composeMiddleware } from '@pins/express';
-import { param } from 'express-validator';
+import { param, query } from 'express-validator';
 import { validationErrorHandler } from '../../middleware/error-handler.js';
 
+/**
+ * @param {string} value
+ * @returns {string}
+ */
+const hasValue = (value) => value;
```
beejjacobs, line 9: It'd be clearer if this returned a boolean, using e.g. `=> Boolean(value)`, since based on its usage we don't need the string return value.

beejjacobs (dismissed):
One minor comment, but otherwise looks good. Good test coverage.

PR #895, opened 2023-05-08
feat(web): add page to upload document version 

shivanshuit914:
## Describe your changes
 - Added page to upload new version of document
 - Added API to upload single document 
 - Added view for document versions

- [BOAS-783](https://pins-ds.atlassian.net.mcas.ms/browse/BOAS-783)

## Type of change 🧩

- [ ] Documentation update

[BOAS-783]: https://pins-ds.atlassian.net/browse/BOAS-783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/server/applications/application/documents/__tests__/document.service.test.js
+++ b/apps/api/src/server/applications/application/documents/__tests__/document.service.test.js
@@ -39,7 +39,7 @@ describe('Document service test', () => {
 		);
 	});
 
-	test('obtainURLForDocumentVersion throws error when document not exist', async () => {
+	test('obtainURLForDocumentVersion uploads new version of document', async () => {
 		databaseConnector.case.findUnique.mockResolvedValue(application);
 		databaseConnector.document.findUnique.mockResolvedValue(document);
 		got.post.mockReturnValue({


apps/api/prisma/migrations/20230509134331_add_document_latest_version_column/migration.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[Document] ADD [latestVersionId] INT CONSTRAINT [Document_latestVersionId_df] DEFAULT 1;
```
DavidJWhiteDev, line 6: I note that all existing documents will have a NULL latestVersionId, and therefore the 2nd version gets numbered as 1.  Do we need to update all existing docs to have latestVersionId = 1 if they are NULL?

DavidJWhiteDev (commented):
when I tested this, it created a new document in the folder, rather than being a latest version on an existing document

apps/api/src/database/migrations/20230509134331_add_document_latest_version_column/migration.sql
```diff
@@ -0,0 +1,19 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[Document] ADD [latestVersionId] INT CONSTRAINT [Document_latestVersionId_df] DEFAULT 1;
```
chrisdalykinandcarta, line 6: This shouldn't default to 1 - let's figure out if we can make it non-nullable and assign it within a transaction when creating the Document and DocumentVersion initially.

apps/api/src/database/schema.prisma
```diff
@@ -387,6 +387,7 @@ model Document {
   fileType             String?
   isDeleted            Boolean           @default(false)
```
chrisdalykinandcarta, line 388: The following properties will be version specific:

- blobStorageContainer
- blobStoragePath
- name (assuming this is display name)
- status
- redacted
- fileSize
- fileType
- representationId

apps/api/src/server/applications/application/documents/document.controller.js
```diff
@@ -53,6 +54,33 @@ export const provideDocumentUploadURLs = async ({ params, body }, response) => {
 	});
 };
 
+/**
+ *
+ * @type {import('express').RequestHandler<any, any, { blobStorageHost: string, blobStorageContainer: string, documents: { documentName: string, blobStoreUrl: string }[] } | any, any>}
+ */
+export const provideDocumentVersionUploadURL = async ({ params, body }, response) => {
+	const documentsToUpload = body;
```
chrisdalykinandcarta, line 62: Is this right that we're supporting providing multiple documents? I think new versions will only be uploaded one at a time

apps/api/prisma/migrations/20230509134331_add_document_latest_version_column/migration.sql
DavidJWhiteDev: The code adds version number 1 automatically for new documents after this, but existing docs in the database will still have a null value, won't they?  Existing docs with null versionid will need to be updated to "1"

apps/api/src/database/migrations/20230509134331_add_document_latest_version_column/migration.sql
DavidJWhiteDev: @chrisdalykinandcarta should it not default to1?  As when a new doc is created it is always version 1

apps/api/src/server/applications/application/documents/document.routes.js
```diff
@@ -81,6 +83,41 @@ router.post(
 	asyncHandler(provideDocumentUploadURLs)
 );
 
+router.post(
+	'/:id/document/:documentId/version',
```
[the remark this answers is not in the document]
DavidJWhiteDev: I still think the endpoint name should indicate what it will do - eg add-version, or upload-version rather than just "version"?

DavidJWhiteDev (approved):
lgtm!

PR #902, opened 2023-05-10
feat(web): add page template for enter start date confirmation page

james-hill-kc:
## Describe your changes

Nunjucks template for enter start date confirmation page + temporary router code to render the page

BOAT-107

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/appeals/appeal-details/appeal-details.router.js
+++ b/apps/web/src/server/appeals/appeal-details/appeal-details.router.js
@@ -24,4 +24,19 @@ router.route('/:appealId/enter-start-date').get(async (request, response) => {
 	}
 });
 
+router.route('/:appealId/start-date-entered').get(async (request, response) => {
+	const appealDetails = await appealDetailsService
+		.getAppealDetailsFromId(request.params.appealId)
+		.catch((error) => logger.error(error));
+
+	if (appealDetails) {
+		response.render('appeals/appeal/start-date-entered.njk', {
+			appeal: {
+				id: appealDetails?.appealId,
+				reference: appealDetails?.appealReference
+			}
+		});
+	}
+});
+
 export default router;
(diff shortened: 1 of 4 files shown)


apps/web/src/server/appeals/appeal-details/appeal-details.router.js
```diff
@@ -24,4 +24,19 @@ router.route('/:appealId/enter-start-date').get(async (request, response) => {
...
+
+	if (appealDetails) {
+		response.render('appeals/appeal/start-date-entered.njk', {
+			appeal: {
+				id: appealDetails?.appealId,
+				reference: appealDetails?.appealReference
+			}
+		});
+	}
+});
```
shepda, line 40: It makes sense to have a conditional to render the page if `appealDetails` is set, but should we have a fallback render/redirect after the conditional to catch the scenario where it's not set?

What would the user see currently, perhaps a white screen?

I see in the description that it's a temporary router so perhaps this will be covered later.

james-hill-kc: It's a good point. I think ideally we should have an error page designed for this scenario (even if it's a generic one across the service). As you say though, this is a temporary router, just to enable the views to be worked on, so this should be addressed as part of BOAT-103/104 (router/controller tickets)

shepda: :+1:

apps/web/src/server/lib/nunjucks-filters/trim.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Trim whitespace from both ends of a string and return a new string without modifying the original.
+ *
+ * @param {string} stringToTrim – The string to trim.
+ * @returns {string} - The trimmed string
+ */
+export function trim(stringToTrim) {
+	return stringToTrim.trim();
+}
```
shepda, line 9: I might have missed something but just wondering how this custom filter differs from the built in one (https://mozilla.github.io/nunjucks/templating.html#trim)?

shepda (approved):
LGTM

If we're building more of these pages we should consider whether we want to create a common component for the confirmation page.

PR #926, opened 2023-05-18
chore(web/appeals): add example new page controller, router, service to help establish standard

AlexMoulin:
Need input from team to agree on how we structure our new pages so they are the same across the board.
Not planning on merging until we all agree!

This will be done in parts as I imagine there will lots of changes required.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/examples/newPage/example.controller.js
+++ b/apps/web/examples/newPage/example.controller.js
@@ -1,0 +1,101 @@
+import logger from '../../src/server/lib/logger.js';
+import * as exampleService from './example.service.js';
+
+/** @typedef {import('./example.types').Example} Example */
+
+/**
+ * @typedef {object} ViewAllExamplesRenderOptions
+ * @property {Example[]} allExamplesFormatted
+ */
+
+/** @type {import('@pins/express').RenderHandler<ViewAllExamplesRenderOptions>}  */
+export async function viewAllExamples(request, response) {
+	const allExamples = await exampleService.getAllExample();
+
+	// Best to check that we have data, as we likely will not be able to display a page if we don't
+	if (allExamples) {
+		// Question: should we create mapper to contain the code below
+		const allExamplesFormatted = [];
+
+		for (const example of allExamples) {
+			if (example) {
+				const formattedExample = {
+					id: example.id,
+					value: example.value
+				};
+
+				allExamplesFormatted.push({ ...formattedExample });
+			}
+		}
+
+		return response.render('./example.njk', { allExamplesFormatted });
+	}
+
+	return response.render('../../src/server/views/app/404.njk');
+}
+
+/**
+ *
+ * @typedef {object} ViewExampleRenderOptions
+ * @property {Example} exampleFormatted
+ */
+
+/** @type {import('@pins/express').RenderHandler<ViewExampleRenderOptions>}  */
+export async function viewExampleById(request, response) {
+	const example = await exampleService.getExampleById(request.params.id);
+
+	// Best to check that we have data, as we likely will not be able to display a page if we don't
+	if (example) {
+		// Question: should we create mapper to contain the code below
+		const exampleFormatted = {
+			id: example.id,
+			value: example.value
+		};
+
+		return response.render('./example.njk', { exampleFormatted });
+	}
+
+	return response.render('../../src/server/views/app/404.njk');
+}
+
+/**
+ *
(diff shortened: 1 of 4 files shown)


apps/web/examples/newPage/example.controller.js
```diff
@@ -0,0 +1,101 @@
...
+/** @type {import('@pins/express').RenderHandler<ViewAllExamplesRenderOptions>}  */
+export async function viewAllExamples(request, response) {
+	const allExamples = await exampleService.getAllExample();
+
+	// Best to check that we have data, as we likely will not be able to display a page if we don't
+	if (allExamples) {
+		// Question: should we create mapper to contain the code below
+		const allExamplesFormatted = [];
+
+		for (const example of allExamples) {
```
fabfab-kc, line 20: I would definitely introduce a formatter / factory here, depending of what extra processing the data needs before going to the template

apps/web/examples/newPage/example.controller.js
```diff
@@ -0,0 +1,101 @@
...
+		// Question: should we create mapper to contain the code below
+		const exampleFormatted = {
+			id: example.id,
+			value: example.value
+		};
+
+		return response.render('./example.njk', { exampleFormatted });
+	}
+
+	return response.render('../../src/server/views/app/404.njk');
```
fabfab-kc, line 58: We should probably add some shortcuts to common views, like a dictionary with 404, 500, etc. or even have some extension methods (are those possible in js?) such as `response.renderNotFound()`, `response.renderError()`...

fabfab-kc: maybe thinking outside the scope of this PR...

apps/web/examples/newPage/example.router.js
```diff
@@ -0,0 +1,19 @@
+import { Router as createRouter } from 'express';
+import * as controller from './example.controller.js';
```
fabfab-kc, line 2: Are we happy with * imports?  Wouldn't be clearer with named imports such as `import { viewAllExamples, viewExampleById } from controller` or is that another conversation :-) ?

apps/web/examples/newPage/example.controller.js
james-hill-kc: I agree it would be nice to have the views in a dictionary or similar - on front office there's a "view" file in "lib" which handles this (helps with reuse/decoupling)

apps/web/examples/newPage/example.router.js
james-hill-kc: This would be a good one to enforce with an eslint rule once we agree a convention

apps/web/examples/newPage/example.service.js
```diff
@@ -0,0 +1,52 @@
+// Import request: this formats requests to the api
+import { get, patch, post } from '../../src/server/lib/request.js';
+
+/** @typedef {import('./example.types').Example} Example */
+/**
+ * @returns {Promise<Array<Example>>} TODO: change Example to dataType returned by Promise
```
james-hill-kc, line 6: Just checking, is the TODO intended to be left in the example?

james-hill-kc:
This is looking great so far. I think it would be good to add an example mapper as you mentioned earlier.

apps/web/examples/newPage/example.controller.js
fabfab-kc: > We should probably add some shortcuts to common views, like a dictionary with 404, 500, etc. or even have some extension methods (are those possible in js?) such as `response.renderNotFound()`, `response.renderError()`...

Extension methods would need to use prototype inheritance, so are out...

apps/web/examples/newPage/example.controller.js
AlexMoulin: I agree! I will look into it.

apps/web/examples/newPage/example.service.js
AlexMoulin: Yes, I will be adding ToDos so that when it is copied and pasted, you know what needs to be modified.
Or does that sound like a bad idea?

apps/web/examples/newPage/example.router.js
AlexMoulin: Well realistically we would always use all the controllers (otherwise we have controllers that are not used) if that makes sense?

PR #967, opened 2023-05-30
feat(api): update create examination table items logic to add sub folder BOAS-599

shivanshuit914:
## Describe your changes

- Create sub-folder when an examination item gets created

- [BOAS-599](https://pins-ds.atlassian.net.mcas.ms/browse/BOAS-599)

## Type of change 🧩

- [ ] Documentation update

[BOAS-599]: https://pins-ds.atlassian.net/browse/BOAS-599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/server/swagger.js
+++ b/apps/api/src/server/swagger.js
@@ -179,7 +179,7 @@ const document = {
 				displayNameCy: 'Accompanied site inspection Cy'
 			}
 		],
-		ExaminationTimetableItem: [
+		ExaminationTimetableItems: [
 			{
 				id: 1,
 				caseId: 1,
--- a/apps/api/src/server/swagger-output.json
+++ b/apps/api/src/server/swagger-output.json
@@ -2292,7 +2292,7 @@
 				}
 			}
 		},
-		"ExaminationTimetableItem": {
+		"ExaminationTimetableItems": {
 			"type": "array",
 			"items": {
 				"type": "object",


apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.controller.js
```diff
@@ -33,5 +35,34 @@ export const createExaminationTimetableItem = async (_request, response) => {
...
 	const examinationTimetableItem = await exminationTimetableItemsRepository.create(body);
 
+	// find the examination folder id to create subfolder.
+	let examinationFolder = await folderRepository.getFolderByNameAndCaseId(
+		body.caseId,
+		'Examination'
+	);
+
+	if (!examinationFolder) {
+		examinationFolder = await folderRepository.createFolder({
```
chrisdalykinandcarta, line 45: Since Examination is a system folder, I think it's best we actually just throw an Error here because it probably means the case has not been configured properly. I think we should just create system folders in a single place, because it prevents drift.

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.controller.js
```diff
@@ -33,5 +35,34 @@ export const createExaminationTimetableItem = async (_request, response) => {
...
+			displayNameEn: 'Examination',
+			caseId: body.caseId,
+			parentFolderId: null,
+			displayOrder: 100
+		});
+	}
+
+	// in some case we don't have date, use mandatory field end date in that case.
+	const date = examinationTimetableItem.date || examinationTimetableItem.endDate;
```
chrisdalykinandcarta, line 54: I'm maybe misunderstanding but 'date' and 'endDate' should be the same thing - examination items will always have a 'date' which is the 'endDate', but deadline items will also have a 'startDate' (in which case, the 'date' is the 'endDate').

So this logic should just always be taking the 'date' which is mandatory. I'm wondering if the entity has been misconfigured?

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.controller.js
```diff
@@ -33,5 +35,34 @@ export const createExaminationTimetableItem = async (_request, response) => {
...
+	// Create sub folder for the examination timetable item.
+	const folderName = `${format(new Date(date), 'dd MMM yyyy')} - ${examinationTimetableItem.name}`;
+
+	const folder = {
+		displayNameEn: folderName,
+		caseId: body.caseId,
+		parentFolderId: examinationFolder?.id || null,
```
chrisdalykinandcarta, line 61: I don't think this should default to null - we should enforce that the examination folder exists.

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.controller.js
```diff
@@ -33,5 +35,34 @@ export const createExaminationTimetableItem = async (_request, response) => {
...
+		displayOrder: 100
```
chrisdalykinandcarta, line 62: I didn't know we had a display order, we can potentially use this instead of the ISO date!

chrisdalykinandcarta: Ah yeah I've just noticed - we're creating timetable items separately here. We could use the unix timestamp of the date?

apps/api/src/database/migrations/20230601144617_remove_end_date_column_examination_items_table/migration.sql
```diff
@@ -0,0 +1,25 @@
...
+
+  - You are about to drop the column `endDate` on the `ExaminationTimetableItems` table. All the data in the column will be lost.
+
+*/
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[ExaminationTimetableItems] DROP COLUMN [endDate];
```
DavidJWhiteDev, line 12: for consistency, should this be singular, rather than plural ie - ExaminationTimetableItem ?  All our other tables are singular - eg Case, Document etc.

DavidJWhiteDev (changes requested):

apps/api/src/database/migrations/20230601144617_remove_end_date_column_examination_items_table/migration.sql
DavidJWhiteDev: I know it means a few name updates throughout, but it would give expected consistency, so I think it would be worth it!

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.routes.js
```diff
@@ -33,16 +33,16 @@ router.get(
...
         #swagger.parameters['id'] = {
             in: 'path',
 			description: 'Examination timetable item ID',
 			required: true,
 			type: 'integer'
         }
         #swagger.responses[200] = {
-            description: 'List of examination timetable items',
+            description: 'Examination timetable item',
             schema: { $ref: '#/definitions/ExaminationTimetableItemsResponseBody' }
```
DavidJWhiteDev, line 46: the response schema definition is now singular, this needs to change as well: Item not Items

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.routes.js
```diff
@@ -62,12 +62,12 @@ router.post(
             required: true
         }
         #swagger.responses[200] = {
-            description: 'List of examination timetable items',
+            description: 'Examination timetable item',
             schema: { $ref: '#/definitions/ExaminationTimetableItemsResponseBody' }
```
DavidJWhiteDev, line 66: the response schema definition is now singular, also this endpoint returns an array of ExaminationTimetableItem.
description of the endpoint should say that it creates a new exam timetable item record, not a get.
Also the body parameters description is incorrect - it is passed a set of exam timetable item properties

apps/api/src/server/applications/examination-timetable-items/examination-timetable-items.routes.js
DavidJWhiteDev, line 46: the response schema definition is now singular, item not items

apps/api/src/server/swagger.js
```diff
@@ -193,7 +193,7 @@ const document = {
 				endTime: '12:20'
 			}
 		],
-		ExaminationTimetableItemsRequestBody: {
+		ExaminationTimetableItemRequestBody: {
 			caseId: 1,
```
DavidJWhiteDev, line 197: the returned body also includes id

DavidJWhiteDev (changes requested):

apps/api/src/server/swagger.js
```diff
@@ -204,7 +204,7 @@ const document = {
 			endDate: '2023-02-27T12:00:00Z',
 			endTime: '12:20'
 		},
-		ExaminationTimetableItemsResponseBody: {
+		ExaminationTimetableItemResponseBody: {
```
DavidJWhiteDev, line 207: field 'endDate' does not exist anymore, should be removed.

DavidJWhiteDev (changes requested):

DavidJWhiteDev (approved):

PR #1029, opened 2023-06-16
feat(applications): store and publish subscriptions

beejjacobs:
## Describe your changes

- Added a schema for nsip-subscription event, and register-nsip-subscription command
- Added an API endpoint to create subscriptions
- Added an Azure Function to respond to register-nsip-subscription commands (Create & Delete)

[ASB-1511](https://pins-ds.atlassian.net/browse/ASB-1511)

## Type of change 🧩

- [ ] Documentation update

[ASB-1511]: https://pins-ds.atlassian.net/browse/ASB-1511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * @type {import('express').RequestHandler}
+ * @throws {Error}
+ * @returns {Promise<void>}
+ */
+export async function getSubscription(request, response) {
+	const { query } = request;
+
+	const caseReference = String(query.caseReference);
+	const emailAddress = String(query.emailAddress);
```
chrisdalykinandcarta, line 18: Is this necessary? Are the query strings being parsed into something else?

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+	};
+
+	if (body.startDate) {
+		subscription.startDate = new Date(body.startDate);
+	}
+	if (body.endDate) {
+		subscription.endDate = new Date(body.endDate);
+	}
+	if (body.language) {
+		subscription.language = body.language;
```
chrisdalykinandcarta, line 59: Do we need to validate this?

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+		);
+
+		response.send({ id: res.id });
+	} catch (/** @type {any} */ e) {
+		/** @type {Object<string, string>} */
+		const errors = {};
+		if (e instanceof Prisma.PrismaClientKnownRequestError) {
+			errors.code = e.code;
+			if (e.code === 'P2002') {
+				errors.constraint = 'caseReference and emailAddress combination must be unique';
```
chrisdalykinandcarta, line 78: Overall I think this is a bit of a design smell, and we should maybe be querying explicitly for the subscription before creating it - and then treating database errors and unhandled - but just an opinion.

For the error specifically though, I'm wondering if 'emailAddress {x} is already subscribed' makes more sense

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if (body.endDate) {
+		subscription.endDate = new Date(body.endDate);
+	}
+	if (body.language) {
+		subscription.language = body.language;
+	}
+
+	try {
+		const res = await subscriptionRepository.create(subscription);
```
chrisdalykinandcarta, line 63: Since we're doing this soft deletion with subscriptions, what happens if somebody subscribes, unsubscribes and then subscribes again?

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+		response.status(400).send({ errors });
+	}
+}
+
+/**
+ * @type {import('express').RequestHandler}
+ * @throws {Error}
+ * @returns {Promise<void>}
+ */
+export async function createSubscription(request, response) {
```
chrisdalykinandcarta, line 42: There's a decent amount of stuff going on here beyond parsing the request and sending a response - wondering if it might be cleaner to move it all into a separate service function or something (think this is a pattern we're using in other places)

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+		if (res === null) {
+			response.status(404).send({ errors: { notFound: 'subscription not found' } });
+		} else {
+			response.send(res);
+		}
+	} catch (/** @type {any} */ e) {
+		/** @type {Object<string, string>} */
+		const errors = {};
+		logger.warn(e, 'unhandled error');
+		errors.unknown = e.message || 'unknown';
```
chrisdalykinandcarta, line 32: Maybe this is a pattern we're using elsewhere, but in general we might consider exception contents to be sensitive - and so don't return them outside of the application

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
```diff
@@ -0,0 +1,144 @@
...
+		response.status(400).send({ errors });
+	}
+}
+
+/**
+ * @type {import('express').RequestHandler}
+ * @throws {Error}
+ * @returns {Promise<void>}
+ */
+export async function updateSubscription(request, response) {
```
chrisdalykinandcarta, line 93: Wondering if this might be best just called 'unsubscribe' or 'endSubscription' or something, to reflect the behaviour

apps/api/src/server/applications/subscriptions/subscriptions.js
```diff
@@ -0,0 +1,31 @@
...
+	const payload = {
+		caseReference: subscription.caseReference,
+		emailAddress: subscription.emailAddress,
+		subscriptionType: subscription.subscriptionType
+	};
+	if (subscription.id) {
+		payload.subscriptionId = subscription.id;
+	}
+	if (subscription.startDate) {
+		payload.startDate = subscription.startDate.toISOString();
```
chrisdalykinandcarta, line 20: JSON serialisation should do this for us (i.e. serialise dates to ISO strings)

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: Yes probably a good idea.

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: Yeah I think in other places we do return them (remember the issue around the prisma error being printed down a web page?), but I agree we shouldn't. I will just put in a generic message instead.

apps/api/src/server/applications/subscriptions/subscriptions.js
beejjacobs: Because the JSON serialisation is implicit, I prefer to be explicit with the payload type. Should mean the tests are closer to the real setup too.

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: I've tried to keep this expandable for future, since there will be a management interface on the back-office down the line. Though appreciate it is very limited in scope right now.

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: See https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/40918285c8246918ac22b7257bdaff10fb68a551

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: Yeah that's fair, see https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/eb007c75ee57bc3b39f5cb86e69e42c9089f39d0

apps/api/src/server/applications/subscriptions/subscriptions.controller.js
beejjacobs: I think we update the `PATCH` endpoint to allow setting both `startDate` and `endDate`, and in the function, if we get a `Create` event for a subscription that already exists, we update `startDate=now` and `endDate=null` accordingly. Any thoughts?

chrisdalykinandcarta: Yeah I think turning this into an upsert operation makes most sense to me!

beejjacobs: Have done this change: https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/943468f90c51f26f77517f1f506f953daa26d1cf

apps/api/src/server/applications/subscriptions/subscriptions.routes.js
```diff
@@ -0,0 +1,127 @@
...
+} from './subscriptions.controller.js';
+import {
+	validateCreateSubscription,
+	validateGetSubscription,
+	validateUpdateSubscription
+} from './subscriptions.validators.js';
+
+const router = createRouter();
+
+router.get(
```
Philip-Philippou1, line 16: Is the use only case for the get by case/email so that we can obtain the id for the delete (update) operation?  Fine if it is and probably re-usable in future.

Philip-Philippou1 (dismissed):

beejjacobs:
I decided to rebase to get this ready.

These few commits are latest changes:
* https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/6740a76b51f8ced3354f02aaa05c30727d1bf015
* https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/7f42aefb15cc99e21eed0e767ffb17e2f7ac241e
* https://github.com/Planning-Inspectorate/back-office/pull/1029/commits/b9f58583a4962e185710cabb034a45e864b80896

apps/api/src/server/applications/subscriptions/subscriptions.service.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Create or update a subscription, and publishes the corresponding event.
+ *
+ * @param {any} request
+ * @returns {Promise<{id: number, created: boolean}>}
+ */
+export async function createOrUpdateSubscription(request) {
+	const subscription = prepareInput(request);
+
+	const existing = await subscriptionRepository.findUnique(
```
Philip-Philippou1, line 20: Just an observation but if caseRef or email is passed in uppercase and lowercase we will end up with 2 subscriptions.  I'm not really proposing we enforce caps though, just something to be aware of.

Philip-Philippou1 (approved):
Great stuff, well done.

apps/api/src/server/applications/subscriptions/subscriptions.service.js
beejjacobs: Good catch, I would normally enforce lowercase for emails - since in practice they are case insensitive. 

Given our (back-office) API is just internal, I think we can get away with it for now!

Philip-Philippou1 (approved):

PR #1057, opened 2023-06-21
chore(tooling): don't run prisma-generate with turbo

beejjacobs:
## Describe your changes

Prisma-generate keeps failing in the pipelines now there are two API apps. Changing to run in series, without turbo, to see if that helps.

N/A

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\"",
     "release": "multi-semantic-release --ignore-packages=packages/**",
     "test": "turbo run test",
-    "tscheck": "npm run prisma-generate --workspace=@pins/api && npm run prisma-generate --workspace=@pins/appeals.api  && turbo run tscheck --parallel --filter=!@pins/api --filter=!@pins/appeals.api",
+    "tscheck": "turbo run prisma-generate --concurrency=1 && turbo run tscheck --parallel --filter=!@pins/api --filter=!@pins/appeals.api",
     "web": "npm run dev --workspace=@pins/web",
     "dev:appeals": "turbo run dev --filter=./appeals/* --no-cache --parallel --continue",
     "test:appeals": "turbo run test --filter=./appeals/*",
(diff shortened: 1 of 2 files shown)


fabfab-kc (approved):

package.json
```diff
@@ -28,7 +28,7 @@
     "prepare": "node -e \"try { require('husky').install() } catch (e) {if (e.code !== 'MODULE_NOT_FOUND') throw e}\"",
     "release": "multi-semantic-release --ignore-packages=packages/**",
     "test": "turbo run test --filter=!./appeals/*",
-    "tscheck": "turbo run prisma-generate && turbo run tscheck --parallel --filter=!@pins/api --filter=!@pins/appeals.api",
+    "tscheck": "npm run prisma-generate --workspace=@pins/api && npm run prisma-generate --workspace=@pins/appeals.api  && turbo run tscheck --parallel --filter=!@pins/api --filter=!@pins/appeals.api",
```
chrisdalykinandcarta, line 31: I think npm run test will also run prisma-generate, and these are running sequentially (looks like we're still filtering on this though, so not sure if this helps)

beejjacobs: Yeah good point. Weirdly `turbo` has a `--parallel` flag, but even without that prisma-generate appeared to run in parallel from the logs!

beejjacobs: `tscheck` seems to fail often, so that's why I targeted that first

turbo.json
```diff
@@ -21,9 +21,6 @@
 			"dependsOn": ["^build"],
 			"outputs": []
 		},
-		"prisma-generate": {
```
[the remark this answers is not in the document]
beejjacobs: My intention here was to remote prisma-generate from turbo because we don't ever want turbo to run it as far as I can tell. It also had `cache: false` anyway.

beejjacobs:
OK I did some digging into this.

`--parallel` flag actually means: run everything and ignore the dependency graph
`--concurrency` is used to determine how many tasks are run in parallel, and tasks that don't depend on each other will always run in parallel. 

So I have set concurrency to 1 for prisma-generate, so they should run one after another.

fabfab-kc (approved):

PR #1084, opened 2023-06-26
fix(web): informational text appears for Deadline for close of examin…

richmoh:
…ation

## Describe your changes
Fix for: 

"Informational text appears for Deadline for close of examination (BOAS-787)"

https://pins-ds.atlassian.net/jira/software/projects/BOAS/boards/172?isInsightsOpen=true&selectedIssue=BOAS-871
## Type of change 🧩

- [ ] Documentation update


beejjacobs (approved):

apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
```diff
@@ -54,6 +54,21 @@ export const timetableTemplatesSchema = {
 	}
 };
 
+export const uniqueTimeTableTypes = {
```
lutamKC, line 57: the point of having the templateType field (stuff like "deadline", "no-times", etc) is to avoid hardcoding the examination timetable items in the frontend (stuff like "accompanied blabla", "publication of", etc).

This list you wrote here contains information that should come from the API `GET applications/examination-timetable-type` and ideally the frontend should be able to work with whatever the result of that API is. For example we should be able to handle timetable types being added, removed or renamed.

richmoh: @lutamKC Understood. The problem I found in this case was that there are two timetable types with templateType `deadline` and we don't want to display the message for a "Deadline For Close Of Examination".

image

In this case the only unique identifiers to separate these two are either the `id` or one of the name fields since now we cannot use the `templateType`. I opted for the `id` since it's possible down the line the names could change. 

What would you suggest as alternative solution?

beejjacobs: Do we need the API to specify this somehow? Is it a difference between deadline types? 

We shouldn't rely on IDs being the same. Sorry I hadn't got the context so perhaps shouldn't have reviewed!

lutamKC (approved):

apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
francescobrenci-AR: we can go with @richmoh current solution then - going to resolve this conversation

francescobrenci-AR (approved):

PR #1108, opened 2023-06-30
feat(api): finish document publishing

chrisdalykinandcarta:
## Describe your changes

Implement file publishing

## Type of change 🧩

- [ ] Documentation update


francescobrenci-AR (dismissed):

apps/functions/document-publish/publish-document/index.js
```diff
@@ -1,6 +1,57 @@
...
+	context.log(`Deploying to file name ${publishFileName}`);
+
+	await blobClient.copyFile({
+		sourceContainerName: config.BLOB_SOURCE_CONTAINER,
+		sourceBlobName: documentURI,
+		destinationContainerName: config.BLOB_PUBLISH_CONTAINER,
+		destinationBlobName: publishFileName
+	});
+
+	await got
```
richmoh, line 34: @chrisdalykinandcarta What would happen if this request failed for whatever reason? Should we consider wrapping in a try catch and reversing any changes made above or does it not matter?

apps/api/src/server/applications/documents/documents.service.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {string} newStatus
+ */
+export const updateStatus = async (guid, newStatus) => {
+	const document = await documentRepository.getByDocumentGUID(guid);
+
+	const currentDocumentStatus = document?.documentVersion[0]?.publishedStatus;
+
+	/** @type {string} */
+	// @ts-ignore
+	const status = nextStatusInDocumentStateMachine(currentDocumentStatus, newStatus);
```
richmoh, line 20: call `nextStatus` ?

apps/functions/document-publish/publish-document/index.js
```diff
@@ -1,6 +1,57 @@
+import config from './config';
```
beejjacobs, line 1: Given the issue I had, this might need `.js`

richmoh (dismissed):
Nice. Approved with a couple of small comments.

packages/blob-storage-client/src/blob-storage-client.js
```diff
@@ -144,20 +143,19 @@ export class BlobStorageClient {
 
 	/**
 	 *
-	 * @param {{currentContainer: string, currentFilePath: string, desiredContainer: string, desiredFilePath: string}} blobStorageHost
+	 * @param {{sourceContainerName: string, sourceBlobName: string, destinationContainerName: string, destinationBlobName: string}} blobStorageHost
```
beejjacobs, line 146: These are better names!

beejjacobs (approved):
Not much to add, looks good.

apps/api/src/server/applications/documents/documents.service.js
chrisdalykinandcarta: It's a good call, but it's passed as a named parameter on the next line and the parameter is called 'status', so just keeping consistent with the repository parameter (but also so I don't have to rename it when passing)

apps/functions/document-publish/publish-document/index.js
chrisdalykinandcarta: Good catch (pun intended), we don't really have a pattern for this at the moment - but if the request fails, we're probably going to assume it's a transient error and bubbling the exception back up to the function runtime will attempt a retry with exponential back-off.

There will potentially be cases where we hit issues we'll want to step back from, just not sure what these are yet. The function itself has to be idempotent so we probably do want to figure out what happens if a document is already in a certain state.

apps/functions/document-publish/publish-document/index.js
chrisdalykinandcarta: Good spot!

francescobrenci-AR (approved):

PR #1134, opened 2023-07-05
feat(appeals) documents upload/download

fabfab-kc:
## Describe your changes

Initial implementation of file upload, download and versioning.
Todo: agree a way for the appeals api to list documents in a specific location.

## Type of change 🧩

- [ ] Documentation update


appeals/api/src/server/endpoints/documents/documents.service.js
```diff
@@ -33,12 +52,143 @@ export const getDocumentsForAppeal = async (appealId, sectionName = null) => {
...
+	return document;
+};
+
+/**
+ * @param {DocumentApiRequest} upload
+ * @param {number} appealId
+ * @returns {Promise<{documents: BlobInfo[]}>}}
+ */
+export const addDocumentsToAppeal = async (upload, appealId) => {
+	const appeal = await getAppeal(appealId);
```
shepda, line 71: There is a middleware function that can be added to a route which uses the `appealId` param and checks that the appeal exists and if so then adds the appeal to the request so it can be used later on.

By using this you shouldn't then need to have an additional method to fetch the appeal separately as you'll already have those details available.

appeals/api/src/server/endpoints/documents/documents.service.js
```diff
@@ -33,12 +52,143 @@ export const getDocumentsForAppeal = async (appealId, sectionName = null) => {
...
 };
+
+/**
+ *
+ * @param {number} caseId
+ * @param {string} reference
+ * @param {DocumentMetadata[]} documents
+ * @returns {Promise<Document[]>}
+ */
+const upsertDocumentMetadata = async (caseId, reference, documents) => {
```
shepda, line 154: There are 3 db interactions within this method, would it make the code more reliable if they were placed in a db transaction so that if any fail then no db updates are done and the db isn't left with a partial update?

appeals/api/src/server/endpoints/documents/documents.service.js
```diff
@@ -33,12 +52,143 @@ export const getDocumentsForAppeal = async (appealId, sectionName = null) => {
...
+	const { results } = await PromisePool.withConcurrency(5)
```
shepda, line 155: Would `Promise.all()` suffice here rather than using `PromisePool`?

fabfab-kc: The concurrency here is to handle multiple uploads in parallel... not sure what's the best way, this is lifted from applications

shepda: `Promise.all()` would do a similar thing as it'll kick off all the promises at the same time then wait until they're all done before it returns.

I guess consistency is also a good thing too though.

fabfab-kc:
Moved to https://github.com/Planning-Inspectorate/back-office/pull/1142

PR #1163, opened 2023-07-11
feat(web): BOAT-240 national list search

kandcdev:
## Describe your changes
national list search
https://pins-ds.atlassian.net/browse/BOAT-240
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/views/appeals/all-appeals/dashboard.njk
+++ b/appeals/web/src/server/views/appeals/all-appeals/dashboard.njk
@@ -1,14 +1,45 @@
-{% from "govuk/components/pagination/macro.njk" import govukPagination %}
+{%- from "govuk/components/pagination/macro.njk" import govukPagination -%}
+{%- from "govuk/components/input/macro.njk" import govukInput -%}
+{%- from "govuk/components/button/macro.njk" import govukButton -%}
+{%- from "govuk/components/table/macro.njk" import govukTable -%}
+
 {%- extends "app/layouts/app.layout.njk" -%}
 
 {%- from "../components/appeals-table.njk" import appealsTable -%}
 
 {%- set containerSize = 'xl' -%}
-{%- set pageTitle = 'Appeals' -%}
+{%- set pageTitle = 'National List' -%}
 
 {%- block pageContent -%}
 	<span class="govuk-caption-xl">{{userRole}}</span>
-	<h2 class="govuk-heading-l govuk-!-margin-bottom-3">All cases</h2>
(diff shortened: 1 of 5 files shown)


appeals/web/src/server/views/appeals/all-appeals/dashboard.njk
```diff
@@ -1,14 +1,46 @@
...
 	<span class="govuk-caption-xl">{{userRole}}</span>
-	<h2 class="govuk-heading-l govuk-!-margin-bottom-3">All cases</h2>
+	<div>
+		<h3 class="govuk-heading-m govuk-!-margin-bottom-3">Search cases</h3>
+		<div class="govuk-caption-m govuk-!-margin-bottom-3">
+			Enter appeal ID or postcode
+		</div>
+		<div>
+			<form action="" method="GET" novalidate>
+				{{ govukTable({
```
james-hill-kc, line 22: If you're using the table component just to get the inline layout of search field and button that's shown in the design, you may be able to avoid this by using a utility class instead. I haven't used any of the custom styling in the codebase yet, but it looks like there are utility classes for display properties eg. `.display--flex` (in `src/styles/4-generic/_generic.display.scss`). There's no issue with using the table, just better to avoid the extra HTML if possible.

james-hill-kc (commented):
Looks good! Just one comment from me

appeals/web/src/server/views/appeals/all-appeals/dashboard.njk
kandcdev: fair enough, but there is a benefit of using govukTable template:
1. I don't need other styling for alignment and spacing of input field and button.
2. It is better to use govuk templates and not using any custom styling

james-hill-kc: Those are fair points. You could probably use the pre-existing govuk utility classes to get the spacing you need (eg. `govuk-!-margin-right-4` - those are found in `/node_modules/govuk-frontend/govuk/overrides/_spacing.scss`). On the second point, in general I agree we should avoid adding new custom styles if we can avoid it, but considering all the styling you need is pre-existing, you wouldn't be adding any. So in this case I think it's fine.

shepda: I agree with James, ideally it would be better not to use the table for layout and use styling instead

kandcdev: Done. Used Grid system instead

shepda (approved):
LGTM

james-hill-kc (approved):

PR #1170, opened 2023-07-12
feat(appeals): adding doc services on web

fabfab-kc:
Introducing a document service in WEB, to validate documents and version uploads. Refactoring and improvements.

https://pins-ds.atlassian.net/browse/BOAT-294

## Type of change 🧩

- [ ] Documentation update


appeals/api/src/server/endpoints/documents/documents.middleware.js
```diff
@@ -1,20 +1,33 @@
...
+
+/**
+ * @type {import("express").RequestHandler}
+ * @returns {Promise<object | void>}
+ */
+export const ensureCaseFolders = async (req, res, next) => {
+	const { appeal } = req;
+	if (!appeal || !appeal.id || !appeal.reference) {
+		return res.status(404).send({ errors: { appealId: ERROR_NOT_FOUND } });
+	}
```
shepda, line 29: The `checkAppealExistsAndAddToRequest` method will check if the appeal exists for the given `appealId` and return an error if not so it shouldn't be necessary to add an additional check here.

fabfab-kc: Yeah, I'll remove though needed to make sure the appeal.reference is there. I suppose it will always be...

shepda: Yes I think there should always be a reference.

However if we do need a specific check for the reference then perhaps we could add it to the generic function so it would then also check other requests.

shepda (approved):
LGTM

PR #1176, opened 2023-07-12
feat(web/applications): create s51 pages (boas-981)

lutamKC:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/applications.router.js
+++ b/apps/web/src/server/applications/applications.router.js
@@ -15,8 +15,13 @@ const domainRouter = createRouter({ mergeParams: true });
 router.use(filters.registerFilters);
 router.use(locals.registerLocals);
 router.use((request, _, next) => {
+	// destroy session containing data for the S51 advice creation journey
+	// triggers in every page
+	// only activates when there is some data in the session
 	const { originalUrl, session } = request;
-	if (!originalUrl.includes('s51-advice/create')) {
+	/** @type{any} */
+	const sessionWithS51 = session;
+	if (sessionWithS51.s51 && !originalUrl.includes('s51-advice/create')) {
 		destroySessionS51(session);
 	}
 	next();
--- a/apps/web/src/server/applications/applications.router.js
+++ b/apps/web/src/server/applications/applications.router.js
@@ -7,12 +7,20 @@ import * as locals from './applications.locals.js';
 import applicationsCaseRouter from './case/applications-case.router.js';
 import applicationsCreateRouter from './create-new-case/applications-create.router.js';
 import applicationsSearchRouter from './search-results/applications-search.router.js';
+import { destroySessionS51 } from './case/s51/applications-s51.session.js';
 
 const router = createRouter();
 const domainRouter = createRouter({ mergeParams: true });
 
 router.use(filters.registerFilters);
 router.use(locals.registerLocals);
+router.use((request, _, next) => {
+	const { originalUrl, session } = request;
+	if (!originalUrl.includes('s51-advice/create')) {
+		destroySessionS51(session);
+	}
+	next();
+});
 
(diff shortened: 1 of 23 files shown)


apps/web/src/server/applications/applications.router.js
```diff
@@ -7,12 +7,20 @@ import * as locals from './applications.locals.js';
...
 
 router.use(filters.registerFilters);
 router.use(locals.registerLocals);
+router.use((request, _, next) => {
+	const { originalUrl, session } = request;
+	if (!originalUrl.includes('s51-advice/create')) {
+		destroySessionS51(session);
+	}
+	next();
+});
```
lutamKC, line 23: PROS: this makes sure to erase the session whenever out of the s51 creation process

CONS: this bit of code is executed for _every_ page in the applications scope

apps/web/src/server/applications/case/documentation/applications-documentation.guard.js
```diff
@@ -0,0 +1,14 @@
...
+ */
+export const assertFolderIsNotReps = (request, response, next) => {
+	if (request.params.folderName === 'relevant-representations') {
+		return response.redirect(
+			`/applications-service/case/${request.params.caseId}/relevant-representations`
+		);
+	}
+
+	next();
+};
```
lutamKC, line 14: not related to the task. Just some code from a RelevantReps ticket that I moved to a more appropriate place (@DavidJWhiteDev )

apps/web/src/server/applications/case/s51/applications-s51.controller.js
```diff
@@ -0,0 +1,67 @@
...
+	'advice-details': { nextPage: 'check-your-answers' }
+};
+
+/**
+ * Show pages for creating/editing s51 advice
+ *
+ * @type {import('@pins/express').RenderHandler<{values: Partial<S51Advice> | null}, {}, {}, {}, {step: string}>}
+ */
+export async function viewApplicationsCaseS51Folder(request, response) {
+	response.render(`applications/case-documentation/folder/documentation-folder`);
```
lutamKC, line 22: this page can be problematic, but it's not part of the ticket, so we can worry about it later.

apps/web/src/server/views/applications/case-documentation/folder/documentation-folder.njk
```diff
@@ -36,56 +40,16 @@
...
-								})
-							}
-						}]
-					}) }}
-				</div>
-
-				{% set userIsInspector = domainType === 'inspector' %}
+				{{ folderHeader(currentFolder, subFolders, caseId) }}
 
-<div class="govuk-width-container">
```
lutamKC, line 62: as I said in the previous comment: the S51 page can be complicated, because it looks like a "documentation-folder" but it's not.

For this task I was able to completely re-use the nunjucks, just adding a few IFs (i.e. if it is S51 show the button for the s51 creation; if it documentation show the button for the Upload).

The part that comes after (actions and table with the items) is going to be more complicated to re-use it and maybe we're going to need to create a whole new template.
For now, this should be fine.

apps/web/src/server/applications/applications.router.js
francescobrenci-AR: maybe check first if s51 data is in the session first - then do the rest if there is related data

apps/web/src/server/views/applications/case-documentation/folder/documentation-folder.njk
francescobrenci-AR: yeah I agree - you can probably abstract from the original the page few component and then create two distinct pages: s51 and documentation pages; and import the common components as it's needed on each

francescobrenci-AR (approved):

PR #1180, opened 2023-07-13
feat(web): appellant case incomplete decision

james-hill-kc:

BOAT-235

## Type of change 🧩

- [ ] Documentation update


appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
```diff
@@ -0,0 +1,230 @@
...
+/** @type {import('@pins/express').RequestHandler<Response>} */
+export const postUpdateDueDate = async (request, response) => {
+	const { errors, body } = request;
+
+	if (errors) {
+		return renderUpdateDueDate(request, response);
+	}
+
+	if (!objectContainsAllKeys(request.session, 'appellantCaseReviewOutcome')) {
+		console.log("!objectContainsAllKeys(request.session, 'appellantCaseReviewOutcome')");
```
beejjacobs, line 183: We should be using pino for logging

beejjacobs, line 183: We should be using pino for logging (if these are meant to be kept in)

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
```diff
@@ -0,0 +1,230 @@
...
+		return response.render('app/500.njk');
+	}
+	if (!objectContainsAllKeys(body, ['due-date-day', 'due-date-month', 'due-date-year'])) {
+		console.log(
```
beejjacobs, line 187: As above, we should be using pino for logging

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
```diff
@@ -0,0 +1,230 @@
...
+		const updatedDueDateDay = parseInt(body['due-date-day'], 10);
+		const updatedDueDateMonth = parseInt(body['due-date-month'], 10);
+		const updatedDueDateYear = parseInt(body['due-date-year'], 10);
+
+		if (
+			Number.isNaN(updatedDueDateDay) ||
+			Number.isNaN(updatedDueDateMonth) ||
+			Number.isNaN(updatedDueDateYear)
+		) {
+			console.log(
```
beejjacobs, line 203: As above, we should be using pino for logging

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-invalid/outcome-invalid.controller.js
```diff
@@ -0,0 +1,130 @@
...
+		const appealReferenceFragments = appealReference.split('/');
+
+		return response.render('appeals/appeal/appellant-case-invalid-incomplete.njk', {
+			appeal: {
+				id: appealId,
+				shortReference: appealReferenceFragments?.[appealReferenceFragments.length - 1]
+			},
+			notValidStatus: 'invalid',
+			reasonOptions: mappedInvalidReasonOptions,
+			otherReasonId: invalidReasonOptions.find((reason) => reason.name.toLowerCase() === 'other')
```
beejjacobs, line 48: This "find 'other'" logic is repeated, do we need a little util function?

appeals/web/src/server/appeals/appeal-details/appellant-case/appellant-case.service.js
```diff
@@ -0,0 +1,70 @@
...
+	return get(`appeals/appellant-case-incomplete-reasons`);
+}
+
+/**
+ * @param {import('./appellant-case.service.js').AppellantCaseReviewOutcome} appellantCaseReviewOutcome
+ * @returns {Promise<import('appeals/api/src/database/schema.js').AppellantCaseInvalidReason[]|import('appeals/api/src/database/schema.js').AppellantCaseIncompleteReason[]>}
+ */
+export async function getAppellantCaseNotValidReasonsForOutcome(appellantCaseReviewOutcome) {
+	switch (appellantCaseReviewOutcome.validationOutcome) {
+		case 'invalid':
```
beejjacobs, line 63: We've got these strings dotted around, seems like a good time to put in a constant file to reference by name.

appeals/web/src/server/appeals/appeal-details/appellant-case/appellant-case.mapper.js
```diff
@@ -19,41 +19,43 @@ export function mapResponseToSummaryListBuilderParameters(appellantCaseData) {
...
 
-	switch (reviewOutcome.validationOutcome) {
-		case 'invalid':
-			if (!reviewOutcome.invalidReasons) {
-				throw new Error('validationOutcome is "invalid" but invalidReasons was not found');
+	(reviewOutcome.validationOutcome === 'invalid'
+		? reviewOutcome.invalidReasons
+		: reviewOutcome.incompleteReasons
+	)?.forEach((reason) => {
+		for (const possibleReason of invalidOrIncompleteReasonOptions) {
```
beejjacobs, line 47: I think this logic would be clearer with a .filter().map call, rather than pushing to an array in a forEach.

beejjacobs (changes requested):
I can't speak much to the domain logic, but have made a few suggested changes.

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
james-hill-kc: Ah, good spot, I left these in by mistake

james-hill-kc: This has been removed now

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
james-hill-kc: This has been removed now

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-incomplete/outcome-incomplete.controller.js
james-hill-kc: This has been removed now

appeals/web/src/server/appeals/appeal-details/appellant-case/outcome-invalid/outcome-invalid.controller.js
james-hill-kc: I've moved this into a util function

appeals/web/src/server/appeals/appeal-details/appellant-case/appellant-case.service.js
james-hill-kc: Moved these into `appeals/web/src/server/appeals/appeal.constants.js`

appeals/web/src/server/appeals/appeal-details/appellant-case/appellant-case.mapper.js
james-hill-kc: I've updated this to use map and filter, personally I find it less clear this way, but I suppose it's a matter of opinion

beejjacobs: I agree with you, I perhaps misunderstood the logic. I thought a filter could be used to remove reasons without a reason option, but it looks like it needs to error instead. I do think the map makes the intention clearer, that each reason will map to an option.

This is a possible option to make it more readable, interested in what you think:

```
const otherNotValidReasons = reviewOutcome.otherNotValidReasons || '';
const reasons = reviewOutcome.validationOutcome === appellantCaseReviewOutcomes.invalid
	? reviewOutcome.invalidReasons
	: reviewOutcome.incompleteReasons;

const reasonsContent = reasons
	?.map(reason => invalidOrIncompleteReasonOptions.find(option => option.id === reason))
	.map(option => {
		if (!option) {
			throw new Error('invalid or incomplete reason ID was not recognised');
		}
		const name = option.name;
		return name.toLowerCase() === 'other' ? `${name}: ${otherNotValidReasons}` : name;
	}) || [''];
```

No need to change though, it's not a big deal for now.

beejjacobs (approved):

beejjacobs (dismissed):

appeals/web/src/server/lib/appellant-case.js
```diff
@@ -0,0 +1,7 @@
+/**
```
fabfab-kc, line 1: Not sure about an 'appellant-case' lib module. Probably could be generalised, as the invalid reasons are id/name pairs and there will be a lot of those coming from the api, but happy to leave it and maybe just rename it?

package-lock.json
fabfab-kc, line 1: Same comment as in appeals/api/package.json

fabfab-kc (changes requested):
LGTM, but a couple of things we should address, mainly package.json changes.

appeals/api/package.json
```diff
@@ -77,7 +77,7 @@
     "jest-environment-jsdom": "^29.3.1",
     "jest-mock-extended": "^3.0.1",
     "nodemon": "^2.0.18",
-    "prisma": "^4.9.0",
+    "prisma": "^4.13.0",
```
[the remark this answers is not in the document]
fabfab-kc: saying that, haven't run npm ci so may quickly change my feedback :-)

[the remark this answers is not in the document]
beejjacobs: we have left it so BO appeals and applications can have different package versions where needed, so this should work. `@prisma/client` is just further up in the dependencies list.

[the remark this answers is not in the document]
james-hill-kc: I'm not sure how this got updated, I don't recall doing it intentionally. I've rolled this back to the original version now and after an `npm ci` all seems well.

appeals/web/src/server/appeals/appeal-details/appellant-case/appellant-case.controller.js
```diff
@@ -0,0 +1,272 @@
...
+
+			notificationBannerParameters = mapReviewOutcomeToNotificationBannerComponentParameters(
+				reasonOptions,
+				notificationData.validationOutcome,
+				notificationData.invalidOrIncompleteReasons,
+				notificationData.otherNotValidReasons
+			);
+
+			// @ts-ignore
+			delete request.session.appellantCaseReviewOutcomeNotificationData;
```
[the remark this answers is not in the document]
james-hill-kc: It is intentional, but on checking the ticket, there's no mention that the notification should disappear after the first view. That was an assumption on my part. I've removed the code related to displaying this banner now, as I realised it wasn't working as intended. As you mentioned, we really need the review outcome, reasons etc. to be returned from the API with the rest of the appellant case data, because using the session as I was originally isn't going to work. I've left the supporting code used to generate and render the banner, as that will be needed once the API has been updated.

appeals/web/src/server/lib/appellant-case.js
james-hill-kc: Well spotted, I think I was too focused on the specific ticket when writing this. I've updated this to a generic helper function which will work with any id/name pairs, and updated the naming accordingly (also added a unit test)

package-lock.json
james-hill-kc: This should be resolved now I've reverted the package.json to the original version, as mentioned above

fabfab-kc (approved):
LGTM!

PR #1186, opened 2023-07-14
Task/fix build

beejjacobs:
## Describe your changes

The build for appeals is broken, and only worked accidentally. It was using the apps package versions from the lock file instead of the appeals one.

I have also tidied up the docker files to be more consistent.

N/A

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/Dockerfile
+++ b/appeals/api/Dockerfile
@@ -4,20 +4,18 @@ COPY package*.json ./
 
 COPY packages ./packages
 
-COPY appeals/api apps/api
+COPY appeals/api appeals/api
 ENV NODE_ENV=production
 RUN npm ci
 COPY appeals/api/.env.development ./.env
-COPY appeals/api/init.sh ./init.sh
 
-ENV SWAGGER_JSON_DIR=./apps/api/src/server/openapi.json
+# relative to cwd
+ENV SWAGGER_JSON_DIR=./src/server/openapi.json
 ENV DATABASE_URL=$DATABASE_URL
 
-WORKDIR /src/app/apps/api
+WORKDIR /src/app/appeals/api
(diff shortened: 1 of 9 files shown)


appeals/api/Dockerfile
```diff
@@ -4,20 +4,18 @@ COPY package*.json ./
...
 COPY appeals/api/.env.development ./.env
-COPY appeals/api/init.sh ./init.sh
 
-ENV SWAGGER_JSON_DIR=./apps/api/src/server/openapi.json
+# relative to cwd
+ENV SWAGGER_JSON_DIR=./src/server/openapi.json
 ENV DATABASE_URL=$DATABASE_URL
 
-WORKDIR /src/app/apps/api
+WORKDIR /src/app/appeals/api
```
fabfab-kc, line 16: Hey @beejjacobs I had left the internal Docker FS layout the same to minimise the changes in scripts and reuse the startup scripts. Not so sure what's the right answer... This is definitely clearer, if it works!

appeals/web/Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY package-lock.json .
 COPY turbo.json .
 
 COPY packages ./packages
-COPY appeals/web ./apps/web
+COPY appeals/web ./appeals/web
```
fabfab-kc, line 18: As above...

fabfab-kc (approved):

appeals/api/Dockerfile
beejjacobs: It was the directory causing the issue. The lock file saves packages are required by each package, and that is referenced by path. So it was installing those that matched apps/api instead of appeals/api

PR #1188, opened 2023-07-17
feat(appeals): mapping of documents and folders to appellant case

fabfab-kc:
Integrates file upload and download on the appellant case

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/src/server/endpoints/documents/documents.routes.js
+++ b/appeals/api/src/server/endpoints/documents/documents.routes.js
@@ -2,19 +2,25 @@ import { Router as createRouter } from 'express';
 import { asyncHandler } from '#middleware/async-handler.js';
 import { getAppealValidator } from '#endpoints/appeals/appeals.validators.js';
 import { checkAppealExistsAndAddToRequest } from '#endpoints/appeals/appeals.service.js';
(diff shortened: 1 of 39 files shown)


appeals/api/src/server/endpoints/documents/documents.routes.js
```diff
@@ -2,27 +2,34 @@ import { Router as createRouter } from 'express';
...
+	getDocumentValidator,
+	getDocumentsValidator
+} from './documents.validators.js';
 import * as controller from './documents.controller.js';
 
 const router = createRouter();
 
 router.get(
-	'/:appealId/document-locations',
+	'/:appealId/document-location/:folderId',
```
shepda, line 17: It looks like this endpoint is for getting the document folders for an appeal? 

If so then it would probably be more consistent for this endpoint to be `/:appealId/document-folders/:folderId` otherwise the endpoint is suggesting that we're getting document locations by folder id and document locations could be folders but could also be something else.

fabfab-kc: ok changed as suggested

shepda (approved):
LGTM

PR #1257, opened 2023-08-02
feat(api): add reference number, statuses column for s51 advice table

shivanshuit914:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/documents/documents.service.js
+++ b/apps/api/src/server/applications/documents/documents.service.js
@@ -24,7 +24,7 @@ export const updateStatus = async (guid, status) => {
 
 	if (status === 'published') {
 		// @ts-ignore
-		updatePayload.publishedStatus = new Date();
+		updatePayload.datePublished = new Date();
 	}
 
 	const updatedDocument = await documentVersionRepository.updateDocumentStatus(updatePayload);


apps/api/src/database/schema.prisma
```diff
@@ -433,19 +433,22 @@ model DocumentActivityLog {
...
+  lastName        String?
+  enquirer        String?
+  enquiryMethod   String
+  enquiryDate     DateTime
+  enquiryDetails  String
+  adviser         String
+  adviceDate      DateTime
+  adviceDetails   String
+  referenceNumber Int      @default(00001)
+  redactedStatus  String   @default("not_redacted")
```
DavidJWhiteDev, line 449: in DocumentVersion, redactedStatus allows null, and no default value - so only changes if the user explicitly sets it to redacted, or not_redacted.  Should we be consistent and do the same thing here?

apps/api/src/database/migrations/20230802104058_add_status_reference_columns_s51_advice/migration.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN TRY
+
+BEGIN TRAN;
+
+-- AlterTable
+ALTER TABLE [dbo].[S51Advice] ADD [publishedStatus] NVARCHAR(1000) NOT NULL CONSTRAINT [S51Advice_publishedStatus_df] DEFAULT 'not_checked',
+[redactedStatus] NVARCHAR(1000) NOT NULL CONSTRAINT [S51Advice_redactedStatus_df] DEFAULT 'not_redacted',
```
DavidJWhiteDev, line 7: in DocumentVersion, redactedStatus allows null, and no default value - so only changes if the user explicitly sets it to redacted, or not_redacted.  Should we be consistent and do the same thing here?

DavidJWhiteDev (changes requested):

DavidJWhiteDev (dismissed):

DavidJWhiteDev (approved):

PR #1266, opened 2023-08-03
fix(web): keyboard-only nav

beejjacobs:
## Describe your changes

It was not possible to get passed the rich-text editor with the keyboard (tabbing). This has been fixed. 

The solution is not ideal at all, unfortunately the library has very little options exposed, and by default overrides the tab behaviour. After much searching and trial-and-error with the library and it's inner workings, this seemed like the best solution unfortunately (short of using the underling library - `prosemirror` and building the editor ourselves).

[ASB-1689](https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-1689)

## Type of change 🧩

- [ ] Documentation update

[ASB-1689]: https://pins-ds.atlassian.net/browse/ASB-1689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/client/components/html-content-editor/html-content-editor.js
+++ b/apps/web/src/client/components/html-content-editor/html-content-editor.js
@@ -24,7 +24,18 @@ function initHtmlContentEditor() {
 			initialEditType: 'wysiwyg',
 			usageStatistics: false,
 			hideModeSwitch: true,
-			toolbarItems: [['bold', 'link', 'ul']]
+			toolbarItems: [['bold', 'link', 'ul']],
+			events: {
+				keydown(_, ev) {
+					// there is no option not to handle tab, but this is an accessibility issue
+					// we want tab to move to the next tab element, not insert spaces
+					if (ev.key === 'Tab') {
+						// by throwing an error, the other handlers don't run and
+						// tab get handled by the browser
+						throw new Error('no error: override tab handler');
+					}
+				}
+			}
 		});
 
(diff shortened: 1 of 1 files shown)


apps/web/src/client/components/html-content-editor/html-content-editor.js
```diff
@@ -24,7 +24,18 @@ function initHtmlContentEditor() {
 			initialEditType: 'wysiwyg',
 			usageStatistics: false,
 			hideModeSwitch: true,
-			toolbarItems: [['bold', 'link', 'ul']]
+			toolbarItems: [['bold', 'link', 'ul']],
+			events: {
+				keydown(_, ev) {
```
Philip-Philippou1, line 29: Seems reasonable, I think this is an edge case, my only question is if the check on each key press slows down the entry in any way, doubt it would.

Philip-Philippou1 (approved):

apps/web/src/client/components/html-content-editor/html-content-editor.js
beejjacobs: Good thought; the library itself checks every key press for many things (including inserting spaces if you press tab). So I think an extra listener will be negligible.

PR #1276, opened 2023-08-07
Deadline submissions

HadrianHughesKC:
## Describe your changes

Create and set up an Azure function to consume deadline submission events from Service Bus and add them to our blob store & db.

Added a unit test for the happy path.

[BOAS-958](https://pins-ds.atlassian.net/browse/BOAS-958)

## Type of change 🧩

- [ ] Documentation update

[BOAS-958]: https://pins-ds.atlassian.net/browse/BOAS-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/src/server/applications/application/application.controller.js
```diff
@@ -90,6 +90,19 @@ export const getApplicationDetails = async ({ params, query }, response) => {
...
 
+/**
+ * @type {import("express").RequestHandler}
+ * */
+export const queryApplications = async ({ query }, response) => {
+	if (!query.reference) {
+		response.end(404);
+	}
+
+	const application = await getCaseByRef(String(query.reference));
```
chrisdalykinandcarta, line 101: Might be worth returning a 404 here if the application doesn't exist

apps/functions/deadline-submissions/deadline-submission-command/index.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {import('@pins/api/src/message-schemas/commands/new-deadline-submission').DeadlineSubmission} msg
+ */
+export default async function (context, msg) {
+	context.log('Handle new deadline submission', msg);
+
+	const client = BlobStorageClient.fromUrl(storageUrl);
+
+	const destinationBlobName = `${msg.deadline}/${msg.submissionType}/${msg.blobGuid}/${msg.documentName}`;
+
+	const result = await client.copyFile({
```
chrisdalykinandcarta, line 19: Before doing this, I think we should probably validate the message - so make sure that the case exists, the timetable item exists, the line item exists (and they're all associated).

We should also check that the file itself exists on the source container - we might be able to do this by looking at the exceptions that client.copyFile could throw.

Whatever the result, we'll need to broadcast some sort of error message back to the front office.

apps/functions/deadline-submissions/deadline-submission-command/index.js
```diff
@@ -0,0 +1,46 @@
...
+	await api.submitDocument({
+		caseID,
+		documentName: msg.documentName,
+		documentType: properties.contentType ?? 'application/octet-stream',
+		documentSize: properties.contentLength ?? 0,
+		folderID,
+		userEmail: msg.email
+	});
+
+	context.log('Copied blob', result);
```
chrisdalykinandcarta, line 45: One last thing we'll have to do is auto-populate some of the metadata (like filter1).

We can do this if we implement the exam timetable endpoint suggested above, because we'll get all the info we need to update the metadata. There are existing APIs for metadata updates that the front-end uses.

apps/functions/deadline-submissions/deadline-submission-command/index.js
```diff
@@ -0,0 +1,46 @@
...
+		// TODO: Publish failure message to service bus
+		context.log('Copying blob failed', result);
+	}
+
+	const properties = await client.getBlobProperties(uploadsContainer, destinationBlobName);
+
+	const caseID = await api.getCaseID(msg.caseReference);
+	const folderID = await api.getFolderID(caseID, msg.deadline, msg.submissionType);
+
+	await api.submitDocument({
```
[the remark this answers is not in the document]
HadrianHughesKC: Currently it looks like code in the web repo uploads documents to blob storage directly, and then calls the API with a reference to the added blob. So yeah it takes 2 calls (blob storage + documents API) otherwise you end up with an orphaned blob document.

chrisdalykinandcarta (approved):

apps/functions/deadline-submissions/deadline-submission-command/back-office-api-client.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ *
+ * @param {number} caseID
+ * @param {string} caseName
+ * @param {string} lineItem
+ * @returns {Promise<number>}
+ * */
+async function getFolderID(caseID, caseName, lineItem) {
```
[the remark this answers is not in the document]
HadrianHughesKC: It could be tricky since line items aren't in a table of their own, they're just in a JSON structure in the ExaminationTimetableItem table. Are you thinking the folder ID would be stored in the JSON structure with the line items?

apps/functions/deadline-submissions/deadline-submission-command/index.js
HadrianHughesKC: I've added checks in to make sure the case, timetable item and line item all exist. I'll set up handling the case where `copyFile` throws, and publishing error messages to service bus in a separate PR.

apps/functions/deadline-submissions/deadline-submission-command/index.js
HadrianHughesKC: Good shout. I'll set this up in a subsequent PR.

apps/functions/deadline-submissions/deadline-submission-command/index.js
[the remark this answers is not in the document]
chrisdalykinandcarta: I think it's the other way around right? The web repo calls the endpoint to get a URL, the endpoint creates a document and then returns the URL to upload against, and then the client (currently the web app, but in this case your function) uploads against that URL.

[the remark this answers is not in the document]
HadrianHughesKC: I've changed the order and now it copies the source blob to the destination given in the API response.

apps/api/src/server/applications/application/application.routes.js
```diff
@@ -39,6 +40,29 @@ router.post(
...
+        #swagger.description = 'Returns the application by its string reference'
+        #swagger.parameters['reference'] = {
+            in: 'query',
+            description: 'Application reference',
+            required: false,
+            type: 'string'
+        }
+        #swagger.responses[200] = {
+            description: 'Application Details',
+            schema: { id: 1, reference: 'AB0110203', status: 'Pre-Application'}
```
beejjacobs, line 57: We could do with a proper swagger schema for Application - but that's beyond the scope of this.

e.g. subscription example: https://github.com/Planning-Inspectorate/back-office/blob/main/apps/api/src/server/swagger.js#L1463

apps/functions/package.json
beejjacobs, line 1: I think you generated a function in the root `functions` folder, should be within a folder (which it looks to me, so maybe all these `functions/*` files just need removing.

azure-pipelines-build.yml
```diff
@@ -255,6 +257,44 @@ extends:
...
+              )
+            )
+          )
+        dependsOn:
+          - Run Linting & Tests
+        steps:
+          - template: ../steps/node_script.yml
+            parameters:
+              nodeVersion: 16
+              script: npm install $(npm pack $(Build.Repository.LocalPath)/packages/event-client) $(npm pack $(Build.Repository.LocalPath)/packages/platform) --only=prod --workspace=@pins/functions-deadline-submissions
```
beejjacobs, line 284: I think you need to add your other `@pins` dependencies here as extra `npm pack` lines, e.g. `blob-storage-client`

azure-pipelines-deploy.yml
```diff
@@ -117,6 +121,16 @@ extends:
...
                   resourceGroup: $(resourceGroup)
+          - name: Deploy Function - Deadline Submissions
+            artifact:
+              name: function-deadline-submissions
+              sourcePipeline: Back Office Build
+            condition: eq(${{ parameters.deployDeadlineSubmissionsFunction }}, 'true')
+            steps:
+              - template: ../steps/function_app_deploy.yml
+                parameters:
+                  appName: pins-func-deadline-submissions-$(ENVIRONMENT)-$(REGION_SHORT)-001
```
beejjacobs, line 132: based on your infra config (https://github.com/Planning-Inspectorate/infrastructure-environments/blob/main/app/components/back-office-app-services/deadline-submissions-function/function-app.tf#L5) `app_name` and `service_name` both get used, so that's why you see the others repeat.

this would be: `pins-func-deadline-submission-deadline-submission-...`

beejjacobs (changes requested):
Generally looks good - a few bits to check over.

beejjacobs (approved):
LGTM!

beejjacobs (approved):

chrisdalykinandcarta (approved):

PR #1296, opened 2023-08-10
feat(applications): notify subscribers

beejjacobs:
## Describe your changes

This feature implements the email notifications for subscribers, sending out Project Updates as emails. It is larger than I'd like, so a break down is provided.

- API:  list subscriptions with filters for caseReference and type: https://github.com/Planning-Inspectorate/back-office/commit/bee80fef32f119bae1c1c609bea3cf875f42579f 
- API: notification log https://github.com/Planning-Inspectorate/back-office/commit/a3c6a1ba2607ac4344be30c0ec918dc8cde9a501
- Function: most of the other commits are for the function implementation itself 

[ASB-1507](https://pins-ds.atlassian.net/jira/software/c/projects/ASB/boards/166?modal=detail&selectedIssue=ASB-1507)

## Type of change 🧩

- [ ] Documentation update

[ASB-1507]: https://pins-ds.atlassian.net/browse/ASB-1507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/database/schema.prisma
+++ b/apps/api/src/database/schema.prisma
@@ -426,6 +426,22 @@ model ProjectUpdate {
   type              String    @default("general") // general | applicationSubmitted | applicationDecided | registrationOpen
 }
 
+// A log of notifications for project updates
+// this will include a line for each subscription that was notified of an update via email
+model ProjectUpdateNotificationLog {
+  id                   Int      @id @default(autoincrement())
+  // ID of the project update this entry relates to
+  projectUpdateId      Int
+  // ID of the subscription this entry relates to
+  subscriptionId       Int
+  // the date this notification was handled
+  entryDate            DateTime
+  // whether an email was successfully sent
+  emailSent            Boolean
+  // the ID of the Azure function run that handled this entry
+  functionInvocationId String
+}
+
 model DocumentActivityLog {
   id              Int             @id @default(autoincrement())
   documentGuid    String
--- a/apps/api/src/server/repositories/project-update.respository.js
+++ b/apps/api/src/server/repositories/project-update.respository.js
@@ -8,18 +8,45 @@ import BackOfficeAppError from '#utils/app-error.js';
  */
 
 /**
- * List project updates for a particular case
+ * @typedef {Object} ListProjectUpdatesOptions
+ * @property {number} page
+ * @property {number} pageSize
+ * @property {import('@prisma/client').Prisma.ProjectUpdateOrderByWithRelationInput} [orderBy]
+ * @property {number} [caseId] - filter option
+ * @property {string} [status] - filter option
+ * @property {Date} [publishedBefore] - filter option (datePublished < publishedBefore)
+ * @property {boolean} [sentToSubscribers] - filter option
+ */
+
+/**
+ * List project updates for a particular case, with filter options such as by case
  *
- * @param {number} caseId
- * @param {number} pageNumber
- * @param {number} pageSize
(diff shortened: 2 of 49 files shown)


apps/api/src/database/schema.prisma
```diff
@@ -423,6 +423,22 @@ model ProjectUpdate {
...
+  // ID of the project update this entry relates to
+  projectUpdateId      Int
+  // ID of the subscription this entry relates to
+  subscriptionId       Int
+  // the date this notification was handled
+  entryDate            DateTime
+  // whether an email was successfully sent
+  emailSent            Boolean
+  // the ID of the Azure function run that handled this entry
+  functionInvocationId String
```
chrisdalykinandcarta, line 439: Whats's this for? Like a trace id?

beejjacobs: Yeah, will be useful for tracking down corresponding logs. It's from the function context.

apps/api/src/server/repositories/project-update.respository.js
```diff
@@ -8,26 +8,59 @@ import BackOfficeAppError from '#utils/app-error.js';
...
+	if (caseId) {
+		where.caseId = caseId;
+	}
+	if (status) {
+		where.status = status;
+	}
+	if (publishedBefore) {
+		where.datePublished = { lt: publishedBefore };
+	}
+	if (sentToSubscribers) {
```
chrisdalykinandcarta, line 53: Should this have three possible values? True, False and Empty? 

Right now if this is passed in as false, the filter will never be applied - so both will be returned.

beejjacobs: Good catch, needs a `typeof sentToSubscribers !== 'undefined'` check

beejjacobs:
Tests are failing before of a routing issue. I need to figure out the best/correct way of defining the routes that doesn't interfere with the existing ones.

apps/functions/notify-subscribers/notify-subscribers/src/util.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Encrypt a value, used for unsubscribe links
+ *
+ * @see https://github.com/Planning-Inspectorate/applications-service/blob/main/packages/applications-service-api/src/lib/crypto.js
+ *
+ * @param {string} value
+ * @returns {string}
+ */
+export function encrypt(value) {
```
chrisdalykinandcarta, line 46: Can we an Azure encryption service for this, so we don't have to manage encryption keys? Also if we really do need to use our own key, we should put it in key vault.

beejjacobs: Possibly, this is mirroring a Front Office function, since they are the ones decrypting. Wel share a key via key vault.

beejjacobs:
Associated infra: https://github.com/Planning-Inspectorate/infrastructure-environments/pull/344

apps/functions/notify-subscribers/notify-subscribers/src/util.js
chrisdalykinandcarta: Makes sense, sounds like a refactor for another time!

chrisdalykinandcarta (approved):

HadrianHughesKC (approved):

PR #1302, opened 2023-08-11
feat(appeals): apiclient injected in request

fabfab-kc:
Injecting api client in request, so we can pass user information from the session. Implemented usage on a couple of services as an example, before refactoring all the remaining ones.

N/A

## Type of change 🧩

- [ ] Documentation update


appeals/web/src/server/appeals/national-list/national-list.service.js
```diff
@@ -1,16 +1,17 @@
...
  * @param {string} searchParam
  * @returns {Promise<AppealList>}
  */
 export const getAppealsByPage = (
+	apiClient,
 	pageNumber = paginationDefaultSettings.firstPageNumber,
 	pageSize = paginationDefaultSettings.pageSize,
 	searchParam = ''
-) => get(`appeals?pageNumber=${pageNumber}&pageSize=${pageSize}${searchParam}`);
+) => apiClient.get(`appeals?pageNumber=${pageNumber}&pageSize=${pageSize}${searchParam}`).json();
```
fabfab-kc, line 17: Had to put `.json()` here, as TS doesn't seem to understand the client config (responseType: 'json',
	resolveBodyOnly: true)

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
```diff
@@ -0,0 +1,41 @@
...
+	resolveBodyOnly: true
+});
+
+const getInstance = (/** @type {string} */ userId) =>
+	instance.extend({
+		hooks: {
+			beforeRequest: [
+				requestLogger,
+				async (options) => {
+					options.headers.userid = userId;
```
beejjacobs, line 21: Are we going with header? I thought we settled on body param, but either way proves the principle.

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @type {import('express').RequestHandler}
+ * @returns {Promise<object|void>}
+ */
+export const addApiClientToRequest = async (req, res, next) => {
+	const user = authSession.getAccount(req.session);
+	if (!user || !user.homeAccountId) {
+		pino.info(`Unauthenticated user redirected to sign in from '${req.originalUrl}'.`);
+		return res.redirect(`/auth/signin?redirect_to=${req.originalUrl}`);
```
beejjacobs, line 36: This is handled by an auth guard right? Not sure if we should duplicate that or throw an error.

appeals/web/src/server/app/app.router.js
```diff
@@ -40,15 +41,20 @@ router.route('/auth/signout').get(asyncRoute(handleSignout));
 
 router
 	.route('/documents/:caseId/upload')
-	.post(assertGroupAccess(config.referenceData.appeals.caseOfficerGroupId), postDocumentsUpload);
+	.post(
+		assertGroupAccess(config.referenceData.appeals.caseOfficerGroupId),
+		addApiClientToRequest,
```
beejjacobs, line 46: I assume we could put this at an `app` or `router` level so it gets applied to all, without changing every route.

That said, might want it to only apply to certain requests (though could use a router.all and filter by method or something).

beejjacobs (commented):
Looks like this should work then!

appeals/web/src/server/appeals/national-list/national-list.service.js
beejjacobs: That'll be because the type used everywhere is `import('got').Got` so it doesn't know it's been configured.

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
fabfab-kc: I think David and I would prefer to go with the header, keeping separate...

shepda: I think we discussed calling `userid` something less generic so it doesn't conflict with other uses of `userid` and it's a bit clearer that it doesn't refer to a value in the BO DB.

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
fabfab-kc: Yeah just to catch the nullable user.homeAccountId and avoid warnings...

appeals/web/src/server/app/app.router.js
fabfab-kc: yeah originally was injecting in the appeals routes only, but I thought it makes sense to make it available to every request

appeals/web/src/server/appeals/national-list/national-list.service.js
fabfab-kc: how would I solve it? `import('got').Response`?

fabfab-kc: or importing the instance maybe...

appeals/web/src/server/app/app.router.js
beejjacobs: Yeah makes sense, so you can apply it at the app/router level then

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
beejjacobs: Yeah I just wonder if we should throw an error, as we shouldn't be able to get to this if.

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
beejjacobs: OK no worries, header will be fine. - yeah something more specific like `authUserId` or `azureAdUserId`

appeals/web/src/server/appeals/national-list/national-list.service.js
beejjacobs: yeah I can't think of a specific way, as `got.extend` just returns `Got` type.

beejjacobs: Looks like others have this issue, with no obvious fix: https://github.com/sindresorhus/got/issues/1117

appeals/web/src/server/app/app.router.js
shepda: `apiClient` is pretty generic and used for all API requests so perhaps this could be done as Ben suggested so it doesn't need to be injected into each route?

shepda (approved):
LGTM

PR #1326, opened 2023-08-15
Feature/deadline submission response

HadrianHughesKC:
## Describe your changes

* feat(api/applications): return fromFrontOffice property on documents
* feat(api/applications): allow fetch documents without application ID
The application ID is actually not used by the endpoint's handler, so it
should be removed from the path. Keeping the old endpoint alongside the
new one to avoid breaking existing consumers.
* feat(functions): log the result of the document virus check
* feat(packages/event-client): add Failure and Success event types
* feat(functions): publish a Service Bus message for result of submissions
* chore(functions): fix test mocks by moving modules into client files
* chore(functions): prepare document-check function for publishing events
* feat(api/applications): allow API to return full list of folders for doc
* feat(functions): publish event after virus check for front office docs
* chore(functions): refactor service bus code to reduce duplication
* feat(functions): only publish event from deadline-submissions on failure
* fix(functions): fix unit tests and add a failure case

Publish Service Bus messages when processing a deadline submissions is successful, unsuccessful, or when a virus is detected.

I've updated unit tests. Will need to test properly in dev.

[BOAS-958](https://pins-ds.atlassian.net/browse/BOAS-958)

## Type of change 🧩

- [ ] Documentation update

[BOAS-958]: https://pins-ds.atlassian.net/browse/BOAS-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/api/src/server/applications/application/documents/document.routes.js
```diff
@@ -233,6 +234,26 @@ router.get(
 	asyncHandler(getDocumentProperties)
 );
 
+router.get(
+	'/document/:guid/properties',
```
beejjacobs, line 238: Is there a reason that "document" is singular? generally rest-like APIs use plural names, and this should be consistent with the other endpoints, e.g. `/applications/{id}/documents/{guid}/properties`

Looks like we're inconsistent here, so might be worth making it a separate PR to sort.

packages/event-client/src/event-type.js
```diff
@@ -6,5 +6,7 @@ export const EventType = {
 	Update: 'Update',
 	Delete: 'Delete',
 	Publish: 'Publish',
-	Unpublish: 'Unpublish'
+	Unpublish: 'Unpublish',
+	Success: 'Success',
```
beejjacobs, line 10: Have these new types been agreed with FO or whoever cares about them?

apps/functions/document-check/check-document/back-office-api-client.js
```diff
@@ -16,6 +16,40 @@ const sendRequestToBackOffice = async (documentGuid, machineAction) => {
...
+ * */
+export const getDocumentProperties = async (documentGuid) => {
+	try {
+		const result = await got
+			.get(`https://${config.API_HOST}/applications/document/${documentGuid}/properties`)
+			.json();
+
+		return result;
+	} catch {
+		return null;
```
beejjacobs, line 32: Maybe this is OK, but it could hide an error that we need to see. Perhaps log it?

apps/functions/document-check/check-document/back-office-api-client.js
```diff
@@ -16,6 +16,40 @@ const sendRequestToBackOffice = async (documentGuid, machineAction) => {
...
+ * */
+export const getDocumentFolders = async (documentGuid) => {
+	try {
+		const result = await got
+			.get(`https://${config.API_HOST}/applications/document/${documentGuid}/path`)
+			.json();
+
+		return result;
+	} catch {
+		return null;
```
beejjacobs, line 49: As above.

apps/functions/document-check/check-document/event-client.js
```diff
@@ -0,0 +1,56 @@
...
+	await eventClient.sendEvents(serviceBus.topic, [event], eventType);
+};
+
+/**
+ *
+ * @param {import('@azure/functions').Context} context
+ * @param {string} guid
+ * */
+export const handleNotInfected = async (context, guid) =>
+	handlePublishResult(context, guid, 'SUCCESS', EventType.Success);
```
beejjacobs, line 48: Probably good to have these constants `SUCCESS` and `VIRUS_DETECTED` in some consts somewhere, so we know what they relate to.

apps/functions/document-check/package.json
```diff
@@ -15,6 +15,7 @@
         "@azure/functions": "^1.2.2",
         "@azure/identity": "3.0.1",
         "@pins/blob-storage-client": "^0.0.0",
+        "@pins/event-client": "^0.0.0",
```
beejjacobs, line 18: I'd expect a corresponding package-lock change for this.

apps/functions/deadline-submissions/deadline-submission-command/event-client.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ *
+ * @param {import('@azure/functions').Context} context
+ * @param {{ deadline: string, submissionType: string, blobGuid: string, documentName: string }} _
+ * */
+async function sendEvent(context, { deadline, submissionType, blobGuid, documentName }) {
+	const eventClient = getEventClient(true, context.log, serviceBusHost);
+
+	const event = {
+		status: 'FAILURE',
```
beejjacobs, line 15: Is this always meant to be failure here?

beejjacobs (changes requested):
Got a few comments/suggestions - feel free to push back on any!

packages/event-client/src/event-type.js
HadrianHughesKC: I haven't agreed them with anyone, I can run it by FO though. It looks like generally so far Service Bus has been used just to publish data when something changes in the DB so that other services can get in sync. This work might be the first example of publishing an actual message, so none of the existing event types make sense for it.

apps/functions/deadline-submissions/deadline-submission-command/event-client.js
HadrianHughesKC: Yeah if this runs it must be a failure. We don't want to publish SUCCESS until the virus check function has run and not found anything. So the happy path for this function would have it not publishing anything, but if it fails it publishes a FAILURE event.

apps/functions/document-check/package.json
HadrianHughesKC: In the root package-lock or for this function? It doesn't look like there is a package-lock for any of the functions so far 🤔

beejjacobs: Root package-lock, as it's a monorepo they all use one lock file.

apps/functions/deadline-submissions/deadline-submission-command/event-client.js
beejjacobs: In that case perhaps the function should be named differently? It doesn't seem obvious this is sending a failure event specifically.

packages/event-client/src/event-type.js
beejjacobs: I wonder if we're overloading this type a little, perhaps this success/failure should just be a message property, as it seems quite specific to this use-case. Not quite sure what it's used for though (these event types are separate to the message so that subscribers can filter based on them).

HadrianHughesKC: Yeah I agree, there is a separate property on the actual event body with the status (success/failure/virus detected). The `sendEvents` function from our event client requires an eventType argument though. I could look at updating it so that that argument is optional.

beejjacobs: OK looking through to understand a bit more, I think these should just be `Create|Update` events - effectively this is just an update to a `nsip-exam-timetable-submission` entity, where the `status` has changed, so I think that fits. Would you agree?

HadrianHughesKC: Ok yeah I think that makes sense. I think it should be Create rather than Update though. FO sends us submissions to add to our DB, and we associate them to cases.

beejjacobs (approved):
Looks good!

You may need to change where you use `getEventClient` so that `serviceBusEnabled` is configurable, so tests can use the dummy client (they're failing atm, and that might be why)

beejjacobs (approved):

PR #1366, opened 2023-08-24
feat(api): get publishable representations endpoint

JamesP-KC:
## Describe your changes

ASB-1709

adds new `GET /applications/{id}/representations/publishable` endpoint

returns all representations that are "publishable" - i.e. `status` is `VALID`, or `status` is `PUBLISHED` and `unpublishedUpdates` is `true`

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/documents/document.routes.js
+++ b/apps/api/src/server/applications/application/documents/document.routes.js
@@ -238,7 +238,7 @@ router.get(
 	'/documents/:guid/properties',
 	/*
         #swagger.tags = ['Applications']
-        #swagger.path = '/document/{guid}/properties'
+        #swagger.path = '/documents/{guid}/properties'
         #swagger.description = 'Gets the properties of a single file'
         #swagger.parameters['guid'] = {
                 in: 'path',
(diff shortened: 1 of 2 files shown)


apps/api/src/server/applications/application/representations/publishable/publishable.route.js
```diff
@@ -0,0 +1,43 @@
...
+                                    required: true,
+                                    type: 'integer'
+            }
+            #swagger.responses[200] = {
+                    description: 'Representations',
+                    schema: {
+                            itemCount: 100,
+                            items:  [
+                                    {
+                                            id: 1,
```
beejjacobs, line 26: We seem to have a "representation" defined in a few places, it'd be better to do a definition once and reference it; or at least `/applications/{id}/representations` seems to return the same type.

beejjacobs (dismissed):
This looks good, one comment, and another question: have we considered pagination for this? Could this list be huge?

apps/api/src/server/applications/application/representations/publishable/publishable.route.js
JamesP-KC: I've actually got the properties wrong in the swagger definition here. This new endpoint returns a smaller subset of the properties which `/applications/{id}/representations` returns.

I was in the processing of adding a single definition as you suggested, but given there's no common definition of representation (all the endpoints return different properties, some with related [contacts, etc] data) this might not make sense.

beejjacobs (approved):
LGTM

beejjacobs (approved):

PR #1380, opened 2023-08-29
fix(api): appellant case and lpaq incomplete flows - should not be able to enter past dates

james-hill-kc:
## Describe your changes

API updates to disallow updated due dates which are in the past for lpaq and appellant case incomplete flows

BOAT-445

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/src/server/endpoints/appellant-cases/__tests__/appellant-cases.test.js
+++ b/appeals/api/src/server/endpoints/appellant-cases/__tests__/appellant-cases.test.js
@@ -6,6 +6,7 @@ import {
 	DEFAULT_DATE_FORMAT_DATABASE,
 	DEFAULT_DATE_FORMAT_DISPLAY,
 	DEFAULT_TIMESTAMP_TIME,
+	ERROR_MUST_BE_IN_FUTURE,
 	ERROR_CANNOT_BE_EMPTY_STRING,
 	ERROR_FAILED_TO_SAVE_DATA,
 	ERROR_INVALID_APPELLANT_CASE_VALIDATION_OUTCOME,
@@ -251,7 +252,7 @@ describe('appellant cases routes', () => {
 				);
 
 				const body = {
-					appealDueDate: '2023-07-14',
+					appealDueDate: '2099-07-14',
 					incompleteReasons: [1, 2, 3],
 					validationOutcome: 'Incomplete',
 					otherNotValidReasons: 'Another reason'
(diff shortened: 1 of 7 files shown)


appeals/api/src/server/endpoints/constants.js
```diff
@@ -67,6 +67,9 @@ export const ERROR_VALID_VALIDATION_OUTCOME_REASONS_REQUIRED =
 	'Validation outcome reasons are required when validationOutcome is Incomplete or Invalid';
 export const ERROR_LPA_QUESTIONNAIRE_VALID_VALIDATION_OUTCOME_REASONS_REQUIRED =
 	'Validation outcome reasons are required when validationOutcome is Incomplete';
+export const ERROR_APPEAL_DUE_DATE_MUST_BE_IN_FUTURE = 'Appeal due date must be in the future';
+export const ERROR_LPA_QUESTIONNAIRE_DUE_DATE_MUST_BE_IN_FUTURE =
+	'LPA questionnaire due date must be in the future';
```
shepda, line 72: The API returns an object with the field name as a parameter so it's clear which field the error relates to, for example

```
{
  "errors": {
    "appealDueDate": "Must be in the future"
  }
}
```

so perhaps it would be better to have a single, more generic error message that could be used for both, then if we need to use for a different field in the future the same error could be reused again, something like

`ERROR_MUST_BE_IN_FUTURE = 'Must be in the future'`

appeals/api/src/server/endpoints/appellant-cases/__tests__/appellant-cases.test.js
```diff
@@ -251,7 +251,7 @@ describe('appellant cases routes', () => {
 				);
 
 				const body = {
-					appealDueDate: '2023-07-14',
+					appealDueDate: '2099-07-14',
```
shepda, line 254: Perhaps we could have an additional test to check that the correct error message is returned if `appealDueDate` in the request body is in the past

appeals/api/src/server/endpoints/lpa-questionnaires/__tests__/lpa-questionnaires.test.js
```diff
@@ -273,7 +273,7 @@ describe('lpa questionnaires routes', () => {
 
 				const body = {
 					incompleteReasons: [1, 2, 3],
-					lpaQuestionnaireDueDate: '2023-06-21',
+					lpaQuestionnaireDueDate: '2099-06-22',
```
shepda, line 276: Perhaps we could have an additional test to check that the correct error message is returned if `lpaQuestionnaireDueDate` in the request body is in the past

appeals/api/src/server/endpoints/lpa-questionnaires/lpa-questionnaires.validators.js
```diff
@@ -67,7 +69,23 @@ const patchLPAQuestionnaireValidator = composeMiddleware(
...
-	validateDateParameter('lpaQuestionnaireDueDate'),
+	validateDateParameter(
+		'lpaQuestionnaireDueDate',
+		(
+			/** @type {any} */ value,
+			/** @type {{ req: { body: { validationOutcome: string } } }} */ { req }
+		) => {
+			if (value && !isOutcomeIncomplete(req.body.validationOutcome)) {
+				throw new Error(ERROR_ONLY_FOR_INCOMPLETE_VALIDATION_OUTCOME);
+			}
```
shepda, line 80: We should probably have a test to check that the correct error message is returned if `lpaQuestionnaireDueDate` is in the request body when the outcome is `Valid`

appeals/api/src/server/endpoints/constants.js
james-hill-kc: Good idea, thanks! Updated.

appeals/api/src/server/endpoints/lpa-questionnaires/__tests__/lpa-questionnaires.test.js
james-hill-kc: Done

appeals/api/src/server/endpoints/lpa-questionnaires/lpa-questionnaires.validators.js
james-hill-kc: Done

shepda (approved):
LGTM

PR #1399, opened 2023-08-30
feat(api): publish event when unpublishing representation

JamesP-KC:
## Describe your changes

ASB-1742

A change to the `PATCH /applications/{id}/representations/{repId}/status` endpoint:

When changing the status of a `Representation` from `PUBLISHED` to another value:
1. publish a `Update` event to the `nsip-representations` topic containing the `Representation` details. 
2. also set the `unpublishedUpdates` flag to `false`.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/representations/status/__tests__/patch-status.test.js
+++ b/apps/api/src/server/applications/application/representations/status/__tests__/patch-status.test.js
@@ -1,19 +1,32 @@
 import { jest } from '@jest/globals';
 import { request } from '../../../../../app-test.js';
+import { eventClient } from '#infrastructure/event-client.js';
 
 const { databaseConnector } = await import('../../../../../utils/database-connector.js');
 
 const existingRepresentations = [
 	{
 		id: 1,
-		representationId: 1,
+		caseId: 1,
 		reference: 'BC0110001-2',
 		status: 'VALID',
 		redacted: true,
 		received: '2023-03-14T14:28:25.704Z'
+	},
+	{
+		id: 2,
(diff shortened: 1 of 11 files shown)


JamesP-KC:
> The ACs may need to change for this, a change was requested to not un-publish WITHDRAWN or ARCHIVED reps, that was not highlighted in the ticket.

As discussed, will continue to publish events when changing to withdrawn/archived, and leave the decision of what to show/hide on the Front Office

Updated the event payload to only include the status and representation ID fields, rather than the whole representation (+contacts, attachments, etc)

apps/api/src/server/applications/application/representations/status/__tests__/patch-status.test.js
```diff
@@ -1,19 +1,32 @@
...
+		reference: 'BC0110001-55',
+		status: 'PUBLISHED',
+		redacted: true,
+		received: '2023-08-11T10:52:56.516Z'
 	}
 ];
 
+const expectedUnpublishPayload = {
+	representationId: 2,
+	status: 'PUBLISHED'
```
Philip-Philippou1, line 28: Should this status  be VALID ?

JamesP-KC: Yes you are absolutely right - I was sending the old status not the newly updated one. I've amended the `publishRepresentationStatusUpdate` function to take the new status value as a parameter now

Philip-Philippou1 (approved):

PR #1431, opened 2023-09-05
feat(appeals/api): update appeals api to add additional lpa questionnaire fields

shepda:
## Describe your changes

Update Appeals endpoint to add LPA Questionnaire fields and updating the Listed Building details.

https://pins-ds.atlassian.net/browse/BOAT-490

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/api/src/server/tests/data.js
+++ b/appeals/api/src/server/tests/data.js
@@ -136,6 +136,7 @@ const householdAppeal = {
 		isAffectingNeighbouringSites: true,
 		isCommunityInfrastructureLevyFormallyAdopted: null,
 		isConservationArea: true,
+		isCorrectAppealType: true,
 		isDevelopmentInOrNearDesignatedSites: null,
 		isEnvironmentalStatementRequired: null,
 		isGypsyOrTravellerSite: null,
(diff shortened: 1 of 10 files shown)


appeals/api/src/server/tests/data.js
```diff
@@ -686,13 +687,27 @@ const baseExpectedAppellantCaseResponse = (appeal) => ({
 	}
 });
 
+const document = {
+	folderId: 1,
+	path: 'path/to/document/folder',
```
fabfab-kc, line 692: There is a routine which splits the document path to extract stage (eg. `lpa_questionnaire`) and doc type (eg: `appealStatement`)... Maybe would be better to have a mock path along those lines (eg .`lpa_questionnaire/emergingPlans`).

fabfab-kc (commented):
LGTM! Just a small suggestion on the document data used in tests...

fabfab-kc (dismissed):

appeals/api/src/server/tests/data.js
shepda: Yeah could do that, I was trying to have a generic test document object that could apply to all documents but we could make it a function and pass in the path as a param.

Otherwise we'll need to create separate test data for each document, which was the path I was going down and then created this to avoid that!

fabfab-kc: no worries, not a big deal - just worried that some validator may impact on tests but sounds all good for now!

fabfab-kc (approved):

PR #1444, opened 2023-09-07
feat(appeals/api): update appeals api to enable assigning of case officer and inspector to an appeal

shepda:
## Describe your changes

- Updated the PATCH Appeal API endpoint to enable the assigning of a Case Officer or Inspector to an Appeal. 

  When the API receives the Azure User Id for either the Case Officer or Inspector a check is done to see if the Azure User Id already exists in the Users table. If it does then the existing User Id is linked to the Appeals table for the relevant user and if it doesn't then a new User is created and the new User Id is linked to the Appeals table for the relevant user.

- Updated the GET Appeal API endpoint to return the Case Officer and/or Inspector who have been assigned to an Appeal 
- Updated the DB to enable both Case Officer and Inspector to be assigned to an Appeal and save the Azure User Id of the assigned user
- Refactored the UUID parameter validator so it can be used for URL and body params and also set to required or not

## Type of change 🧩

- [ ] Documentation update


fabfab-kc (approved):
LGTM!

FYI, related... We have a requirement to send the SAP ID of the user in the appeal topic messages that we'll need for integration. The SAP ID will be in the ODW `employee` topic/dataset, which I'm working on importing. Then, we'll need to find a way / event to associate the `azureUserId` and the SAP ID (probably by email match, but the WEB needs to gives us the value...

PR #1466, opened 2023-09-08
feat(web/applications): add key dates page (boas-684)

lutamKC:
## Type of change 🧩

- [ ] Documentation update


apps/web/testing/applications/factory/util.js
```diff
@@ -29,7 +29,7 @@ export const createUniqueRandomNumberFromSeed = (min, max, seed) => {
 	const randomFloat =
 		Number.parseInt([...date.getTime().toString()].reverse().slice(-3).join(''), 10) / 1000;
 
-	return Math.floor(randomFloat * max) + min;
+	return Math.floor(randomFloat * (max - min)) + min;
```
lutamKC, line 32: the reason why kind of every snapshot has changed.

apps/web/src/styles/7-components/ListMenu/ListMenu.scss
lutamKC, line 1: refactored as other files in the format /components/name-of-style.scss

apps/web/src/server/views/applications/case/layouts/applications-case-layout.njk
```diff
@@ -41,13 +41,13 @@
 {% endblock %}
 
 {% block pageContent %}
-	<div class="govuk-grid-row">
+	<div class="govuk-grid-row display--flex">
 		{% block sectionMenu %}
-			<div class="govuk-grid-column-one-third">
+			<div class="pins-list-menu-container">
```
lutamKC, line 46: side menu of case pages (all of them), was shrinked. From 1/3 of the page (standard govuk width), to only 220px.

this is one of the reason of the many updated snapshots

apps/web/src/server/lib/nunjucks-filters/key-dates-property.js
```diff
@@ -0,0 +1,65 @@
+/** @type {Record<string, string>} */
+const propertyToName = {
```
lutamKC, line 2: I hope the mapping is correct. 

Thanks to @chrisdalykinandcarta  the names were straightforwarding, but I could have done some mistake anyway.

apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
```diff
@@ -87,6 +87,7 @@ export async function viewApplicationsCaseTimetableList(_, response) {
 	response.render(`applications/case-timetable/timetable-list`, {
 		timetableItems: timetableItemsViewData,
 		publishedStatus: examinationTimetable?.published,
+		selectedPageType: 'examination-timetable',
```
lutamKC, line 90: unrelated fix: highlight the menu item when on the page

apps/api/src/server/utils/mapping/map-key-dates.js
```diff
@@ -39,24 +39,24 @@ export const mapRequestToKeyDates = ({
...
-			...mapValues(
-				pick(
-					keyDates,
-					preApplicationDateNames.filter((key) => key != 'submissionAtPublished')
-				),
-				mapDateToUnixTimestamp
-			),
-			submissionAtPublished: keyDates.submissionAtPublished
-		},
+		preApplication: mapValues(pick(keyDates, preApplicationDateNames), mapPreApplicationDates),
```
lutamKC, line 59: design expects properties to be in a specific order, at least in this section ("Anticipated submission date published" should be right before "Anticipated submission date internal").

in this way we only map the dates as it was before (so we exclude submissionAtPuslibhshed: string), but keeping the same order of the properties

chrisdalykinandcarta: Makes sense, thanks for explanation! Also I think your code is more readable than mine anyway.

francescobrenci-AR (approved):

PR #1501, opened 2023-09-15
feat(api): implement migration job for project updates

chrisdalykinandcarta:
## Describe your changes

Implemented migration job for project updates. I chose readability over efficiency (e.g. could have pulled back all existing updates in one query using IN, could also just record created entity IDs and do a big SELECT at the end), this is because the volume of updates and subscriptions will be extremely low per case.

## Type of change 🧩

- [ ] Documentation update


packages/event-client/src/service-bus-event-client.js
```diff
@@ -1,6 +1,10 @@
 import { DefaultAzureCredential } from '@azure/identity';
 import { ServiceBusClient } from '@azure/service-bus';
 
+// 100 events would allow each event to have a body size of 10kb which is more than enough
+// I haven't seen an event in the test environment larger than 1kb
+const eventChunkSize = 100;
```
beejjacobs, line 6: I'm wondering if this logic should live where the event-client is used instead of in this package. Then the application has control over chunk size and failure handling. With this implementation you don't know which set of events failed to send. Any thoughts?

apps/api/src/server/migration/migrators/nsip-project-update-migrator.js
```diff
@@ -0,0 +1,184 @@
...
+export const migrateNsipProjectUpdates = async (projectUpdates) => {
+	console.info(`Migrating ${projectUpdates.length} project updates`);
+
+	/** @type {UpdateToBroadcast[]} */
+	const updatesToBroadcast = [];
+
+	for (const model of projectUpdates) {
+		const entity = await mapModelToEntity(model);
+
+		// We can't make migratedId unique because it's nullable, and MSSQL doesn't allow nullable unique fields
```
beejjacobs, line 41: more accurate to say Prisma doesn't allow it - MSSQL does.

apps/api/src/server/migration/migrators/nsip-project-update-migrator.js
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * @param {NSIPProjectUpdateMigrateModel[]} projectUpdates
+ */
+export const migrateNsipProjectUpdates = async (projectUpdates) => {
+	console.info(`Migrating ${projectUpdates.length} project updates`);
+
+	/** @type {UpdateToBroadcast[]} */
+	const updatesToBroadcast = [];
+
+	for (const model of projectUpdates) {
```
beejjacobs, line 38: Do we want to try/catch the logic in the for, so we can still process any that succeed?

apps/api/src/server/migration/migrators/nsip-project-update-migrator.js
```diff
@@ -0,0 +1,184 @@
...
+		// @ts-ignore
+		title: m.updateName,
+		htmlContent: sanitizeHtml(m.updateContentEnglish, {
+			allowedTags,
+			allowedAttributes: {
+				a: ['href']
+			},
+			allowedSchemes: ['https']
+		})
+		// TODO: Do we also need to migrate update type?
```
beejjacobs, line 100: I don't think we can know update type from the old data

beejjacobs (commented):
Got a few comments and questions, but generally looks great. I think the generic model-based API endpoint is a good solution.

apps/api/src/server/migration/migrators/nsip-project-update-migrator.js
chrisdalykinandcarta: So the design is currently "if any fail then they all fail", and we propagate an error back to the front-end. This is just because it makes the control flow easier, since the end-user will have to fix something and since the operation is idempotent they can try everything again.

If any fail to insert, I don't think we should broadcast any events - because the front-end would wind up with a partial list of banners, so I just thought abandoning ship entirely was easier.

What are your thoughts?

packages/event-client/src/service-bus-event-client.js
chrisdalykinandcarta: You're right - hadn't thought about failures. I configured it this way because adding an additional parameter would be difficult since we aren't using an object and there's already a nullable parameter at the end of the parameter list. I'll take this logic out of the client entirely and let the user manage this instead.

Maybe a max events array size is a better idea, and we throw an error?

beejjacobs (approved):
Looks good (changes requested coming in future PR).

PR #1549, opened 2023-09-22
fix(api): ensure required fields are checked before start

beejjacobs:
## Describe your changes

Validate api config before app start - i.e. set fields to required that are required.

[BOAS-1100](https://pins-ds.atlassian.net/browse/BOAS-1100)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1100]: https://pins-ds.atlassian.net/browse/BOAS-1100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/server/config/schema.js
+++ b/appeals/api/src/server/config/schema.js
@@ -1,68 +1,75 @@
 import joi from 'joi';
 
-export default joi.object({
-	NODE_ENV: joi.string().valid('development', 'production', 'test'),
-	PORT: joi.number(),
-	SWAGGER_JSON_DIR: joi.string(),
-	DATABASE_URL: joi.string(),
-	DISABLE_INTEGRATIONS: joi.boolean(),
-	defaultApiVersion: joi.string(),
-	serviceBusOptions: joi.object({
-		hostname: joi.string().optional()
-	}),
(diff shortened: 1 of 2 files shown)


appeals/api/src/server/config/schema.js
```diff
@@ -1,68 +1,75 @@
...
+			.object({
+				api: joi.object({
+					key: joi.string() // optional if not NODE_ENV=production
+				}),
+				template: joi.object({
+					validAppellantCase: joi.object({
+						id: joi.string()
+					})
+				}),
+				testMailbox: joi.string()
```
shepda, line 54: For completeness perhaps `testMailbox` could be optional when not production as with `api.key` as the two are related but it also works fine as it is too.

shepda (approved):

appeals/api/src/server/config/schema.js
beejjacobs: That's fair, the reason I left if its that testMailbox has a default even if TEST_MAILBOX isn't set - so it'll always pass validation.

PR #1556, opened 2023-09-25
Feature/boas 1137 unpublished changes

HadrianHughesKC:
## Describe your changes

* feat(api/applications): add a column for cases with unpublished changes
* feat(api/applications): implement updating hasUnpublishedChanges field
Check if any fields which were updated are fields which will eventually be
published. If so, set `hasUnpublishedChanges` to true.
* feat(api/applications): set hasUnpublishedChanges to false on publish
* feat(api/applications): return hasUnpublishedChanges in response
* feat(web/applications): render banner for unpublished changes
* chore(web/applications): optimistically set hasUnpublishedChanges to false
* test(api/applications): update test for publishing application
* chore(api/applications): compare stringified case events to diff changes

Tested locally.

[BOAS-1137](https://pins-ds.atlassian.net/browse/BOAS-1137)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1137]: https://pins-ds.atlassian.net/browse/BOAS-1137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/server/applications/application/application.controller.js
+++ b/apps/api/src/server/applications/application/application.controller.js
@@ -7,7 +7,7 @@ import logger from '../../utils/logger.js';
 import BackOfficeAppError from '../../utils/app-error.js';
 import { mapCaseStatusString } from '../../utils/mapping/map-case-status-string.js';
 import { mapDateStringToUnixTimestamp } from '../../utils/mapping/map-date-string-to-unix-timestamp.js';
-import { publishedFieldsHaveChanged } from '../../utils/published-fields-changed.js';
+import { setCaseUnpublishedChangesIfTrue } from '../../utils/published-fields-changed.js';
 import { buildNsipProjectPayload } from './application.js';
 import { mapCreateApplicationRequestToRepository } from './application.mapper.js';
 import { getCaseDetails, getCaseByRef, startApplication } from './application.service.js';
@@ -76,22 +76,7 @@ export const updateApplication = async ({ params, body }, response) => {
 		throw new BackOfficeAppError('Application not found', 500);
 	}
 
-	const publishableFieldsChanged = publishedFieldsHaveChanged(originalResponse, updateResponse);
-	if (publishableFieldsChanged) {
-		const finalResponse = await caseRepository.updateApplication({
-			caseId: params.id,
-			hasUnpublishedChanges: true
-		});
-
-		if (!finalResponse) {
-			throw new BackOfficeAppError(
-				`Could not update hasUnpublishedChanges propery for case with id ${params.id}`,
-				500
-			);
-		}
-
-		updateResponse = finalResponse;
-	}
+	updateResponse = await setCaseUnpublishedChangesIfTrue(originalResponse, updateResponse);
 
 	await eventClient.sendEvents(
 		NSIP_PROJECT,
--- a/apps/api/src/server/applications/application/application.controller.js
+++ b/apps/api/src/server/applications/application/application.controller.js
@@ -7,7 +7,7 @@ import logger from '../../utils/logger.js';
 import BackOfficeAppError from '../../utils/app-error.js';
 import { mapCaseStatusString } from '../../utils/mapping/map-case-status-string.js';
 import { mapDateStringToUnixTimestamp } from '../../utils/mapping/map-date-string-to-unix-timestamp.js';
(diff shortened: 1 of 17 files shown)


apps/api/src/server/applications/application/application.controller.js
```diff
@@ -49,17 +50,49 @@ export const createApplication = async (request, response) => {
...
 		applicantId: head(body?.applicants)?.id,
 		...mappedApplicationDetails
 	});
 
 	if (!updateResponse) {
-		throw new Error('Application not found');
+		throw new BackOfficeAppError('Application not found', 500);
+	}
+
+	const publishableFieldsChanged = publishedFieldsHaveChanged(originalResponse, updateResponse);
```
beejjacobs, line 79: Is it worth checking `hasUnpublishedChanges`? If it's already true we don't need to check again.

apps/api/src/server/applications/key-dates/key-dates.controller.js
```diff
@@ -38,9 +39,25 @@ export const updateKeyDates = async ({ body, params }, response) => {
 
 	const keyDates = mapRequestToKeyDates(body);
 
+	const originalCase = await caseRepository.getById(id, {
```
beejjacobs, line 42: This looks similar to `application.controller.js` - are there some service or repo methods we should be adding to hold this logic?

apps/api/src/server/utils/published-fields-changed.js
```diff
@@ -0,0 +1,49 @@
...
+/** @typedef {import('@pins/applications.api').Schema.Case} Case */
+
+/**
+ * Checks whether the given column name is included when the case is published
+ *
+ * @param {Case} original
+ * @param {Case} updated
+ * @returns {boolean}
+ * */
+export const publishedFieldsHaveChanged = (original, updated) => {
```
beejjacobs, line 12: This looks like it'd work fine, but if we change the schema/payload we have to update this too.

What if instead we build an event with the before and after cases, and compare them (i.e. using `buildNsipProjectPayload`)? Could even just stringify and check if they're equal. That way it's tied into what we publish.

beejjacobs (changes requested):
Looks good, just got some suggestions/ideas.

Also, should there be a swagger/api-spec update for this?

apps/api/src/server/utils/published-fields-changed.js
HadrianHughesKC: Yeah good idea, I've made that change. We just need to make sure the `buildNsipProjectPayload` function only includes properties that should be in the event. There's some `@ts-ignore` dotted around that might allow extra properties to be included. I think it's right for now though.

apps/api/src/server/applications/key-dates/key-dates.controller.js
HadrianHughesKC: I've made the change below and done it in a common util so there's no duplication.

apps/api/src/server/utils/published-fields-changed.js
```diff
@@ -0,0 +1,53 @@
...
+/** @typedef {import('@pins/applications.api').Schema.Case} Case */
+
+/**
+ * Checks whether the given column name is included when the case is published
+ *
+ * @param {Case} original
+ * @param {Case} updated
+ * @returns {boolean}
+ * */
+const publishedFieldsHaveChanged = (original, updated) => {
```
beejjacobs, line 14: I think its worth making it clear in the name this is for cases, not generic.

`publishedCaseFieldsHaveChanged` perhaps?

beejjacobs (commented):
Looks good. One minor comment below. Also does this change the API returned value? (i.e. do we include `hasUnpublishedChanges`? If so I'd expected a swagger-output change.

apps/api/src/server/utils/published-fields-changed.js
HadrianHughesKC: That's done

HadrianHughesKC:
> Looks good. One minor comment below. Also does this change the API returned value? (i.e. do we include `hasUnpublishedChanges`?) If so I'd expected a swagger-output change.

Thanks @beejjacobs I've regenerated swagger.

PR #1566, opened 2023-09-27
fix(web/applications): fix exam timetable display no items (boas-1200)

DavidJWhiteDev:
## Describe your changes

Examination timetable page crashes if there are no exam timetable items on that case.  Fixed.

## BOAS-1200 Fix Exam Timetable Display when no items
https://pins-ds.atlassian.net/browse/BOAS-1200

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
+++ b/apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
@@ -80,7 +80,7 @@ export const uniqueTimeTableTypes = {
 export async function viewApplicationsCaseTimetableList(_, response) {
 	const examinationTimetable = await getCaseTimetableItems(response.locals.caseId);
 	const timetableItemsViewData = examinationTimetable?.items?.map(getTimetableRows) ?? [];
-	const republishStatus = examinationTimetable.items.some(
+	const republishStatus = examinationTimetable.items?.some(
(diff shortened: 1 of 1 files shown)


apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
```diff
@@ -80,7 +80,7 @@ export const uniqueTimeTableTypes = {
 export async function viewApplicationsCaseTimetableList(_, response) {
 	const examinationTimetable = await getCaseTimetableItems(response.locals.caseId);
 	const timetableItemsViewData = examinationTimetable?.items?.map(getTimetableRows) ?? [];
-	const republishStatus = examinationTimetable.items.some(
+	const republishStatus = examinationTimetable.items?.some(
```
lutamKC, line 83: the "some" method returns a boolean. the "?" will return undefined rather than false. 
is that ok?

I think it'll work because nunjucks is not that smart to tell difference between false and undefined, but better to be sure.

lutamKC (approved):

apps/web/src/server/applications/case/examination-timetable/applications-timetable.controller.js
DavidJWhiteDev: It passes testing on a case with and without exam items, but I can change this to something else if you prefer?

PR #1588, opened 2023-09-29
Document API changes + BOAS-1204 fix

HadrianHughesKC:
## Describe your changes

* refactor(api/applications): improve code consuming document update API
* feat(api/applications): remove `/update` from path of document update
* refactor(api/applications): improve code structure in update handler
* feat(api/applications): return results + errors when updating documents
* refactor(api/applications): code improvements in document controller
* fix(web/applications): display inline document errors send from API
* chore(api/applications): generate Swagger spec

I did decide to make the changes in one PR as it would have been tricky to separate them out.
The ticket also mentions S51 items---I'll do that next in another PR.

Tested locally.

[BOAS-1204](https://pins-ds.atlassian.net/browse/BOAS-1204)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1204]: https://pins-ds.atlassian.net/browse/BOAS-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/server/applications/application/documents/document.controller.js
+++ b/apps/api/src/server/applications/application/documents/document.controller.js
@@ -159,6 +159,7 @@ export const updateDocuments = async ({ body }, response) => {
 		const { publishable, invalid } = await verifyAllDocumentsHaveRequiredPropertiesForPublishing(
 			documentIds
 		);
+
 		if (invalid.length > 0) {
 			errors = [
 				...errors,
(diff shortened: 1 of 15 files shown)


francescobrenci-AR (approved):

apps/api/src/server/applications/application/documents/document.controller.js
```diff
@@ -140,78 +140,112 @@ export const updateDocuments = async ({ body }, response) => {
...
 	// and the redaction status of each document should remain unchanged.
-	if (typeof isRedacted !== 'undefined') {
-		redactedStatus = getRedactionStatus(isRedacted);
-	}
+	const redactedStatus = isRedacted !== undefined ? getRedactionStatus(isRedacted) : undefined;
 
 	// special case - for Ready to Publish, need to check that required metadata is set on all the files - else error
-	if (publishedStatus === 'ready_to_publish') {
-		const documentIds = documents.map((/** @type {{ guid: string; }} */ document) => document.guid);
+	/** @type {{ publishableIds: string[], errors: { guid: string, msg: string }[] }} */
```
chrisdalykinandcarta, line 148: I have a few thoughts on this endpoint overall, I think it's quite confusing to read (not that you've made it this way, it was always confusing!)

Can we separate the anonymous function to a different call entirely, only call it if the status is being set to 'ready_to_publish', and then just throw an error if there are _any_ documents we can't update? I know this breaks the contract, but it makes the whole thing read quite confusing (e.g. publishableDocuments being used throughout, and just defaulting this to all documents which seems like a workaround for our own code)

HadrianHughesKC: Would it not be better to collect the errors and and return them with the successful results?

I agree about moving the anonymous function out to a named function, I'll make that change. I think it should be possible to improve the readability without sacrificing the API interface we want though.

HadrianHughesKC: I've pushed a change that should make the controller more readable.

chrisdalykinandcarta (approved):

PR #1617, opened 2023-10-05
fix(appeals): fix site-visit

AlexMoulin:
Fixes site visit schedule, prepopulates with exisiting values, saves data, and redirects correctly

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.controller.js
+++ b/appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.controller.js
@@ -3,6 +3,7 @@ import * as appealDetailsService from '../appeal-details.service.js';
 import * as siteVisitService from './site-visit.service.js';
 import {
 	mapWebVisitTypeToApiVisitType,
+	mapGetApiVisitTypeToWebVisitType,
 	buildSiteDetailsSummaryListRows
 } from './site-visit.mapper.js';
 import {
@@ -25,7 +26,7 @@ const renderScheduleSiteVisit = async (request, response) => {
 		.catch((error) => logger.error(error));
 
 	if (appealDetails) {
-		const {
+		let {
 			body: {
 				'visit-type': visitType,
 				'visit-date-day': visitDateDay,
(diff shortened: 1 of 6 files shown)


appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.controller.js
```diff
@@ -38,6 +38,30 @@ const renderScheduleSiteVisit = async (request, response) => {
 			}
 		} = request;
 
+		// Nullish coalescing assignment https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment
+		visitType ??= appealDetails.siteVisit.visitType;
```
james-hill-kc, line 42: Does this mean existing values from the API response will overwrite values in the POST request body? If so I think it should be the other way around

appeals/web/src/server/views/appeals/appeal/schedule-site-visit.njk
```diff
@@ -79,17 +79,17 @@
 				{
 					value: "unaccompanied",
 					text: "Unaccompanied",
-					checked: siteVisit.visitType === "unaccompanied"
+					checked: (siteVisit.visitType === "unaccompanied" or siteVisit.visitType === "Unaccompanied")
```
james-hill-kc, line 82: If this is to handle values from the API response, it should probably be done in the mapper instead

james-hill-kc (changes requested):
Looks good! Just a couple of comments from me

appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.controller.js
AlexMoulin: Yes but only if they are not defined.

appeals/web/src/server/views/appeals/appeal/schedule-site-visit.njk
AlexMoulin: I think the best thing to do would be to fix the API so the Get (Unaccompanied) and Post (unaccompanied) are the same value formats. I'll create a ticket to address that and put a note to remove redundant code after it has been done. Would that be sufficient or should I build a temporary solution in the mapper too?

appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.controller.js
james-hill-kc: Oh I see now, I was misunderstanding how nullish coalescing works

appeals/web/src/server/views/appeals/appeal/schedule-site-visit.njk
james-hill-kc: I agree this should be fixed API side ideally. In principle though there is still mapping occurring directly in the view here, strictly speaking it should be moved into the mapper regardless of what the API values are

james-hill-kc (approved):
Looks good!

PR #1678, opened 2023-10-17
feat(appeals/api): publishes service users

fabfab-kc:
- Introduce publishing to topic for service users (appellants, agents, etc).
- Validation of messages with schemas
- Broadcasts appeal updates to SB as a result of updating appeal data in the BO (still work to do here).

[BOAT-611](https://pins-ds.atlassian.net/browse/BOAT-611)

## Type of change 🧩

- [ ] Documentation update

[BOAT-611]: https://pins-ds.atlassian.net/browse/BOAT-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/api/src/message-schemas/pins-root.schema.json
+++ b/appeals/api/src/message-schemas/pins-root.schema.json
@@ -15,8 +15,6 @@
 				"originalCaseDecisionDate",
 				"costsAppliedForIndicator",
 				"LPAApplicationReference",
-				"appellant",
-				"agent",
 				"siteAddressLine1",
 				"siteAddressLine2",
 				"siteAddressTown",
@@ -53,36 +51,6 @@
 				},
 				"LPAApplicationReference": {
 					"type": "string"
-				},
-				"appellant": {
-					"type": "object",
-					"properties": {
-						"firstName": {
-							"type": "string"
-						},
-						"lastName": {
-							"type": "string"
-						},
-						"emailAddress": {
-							"type": "string"
-						}
-					},
-					"required": ["firstName", "lastName", "emailAddress"]
-				},
-				"agent": {
-					"type": "object",
-					"required": ["firstName", "lastName"],
-					"properties": {
-						"firstName": {
-							"type": "string"
-						},
-						"lastName": {
-							"type": "string"
-						},
-						"emailAddress": {
-							"type": "string"
-						}
-					}
 				},
 				"siteAddressLine1": {
 					"type": "string"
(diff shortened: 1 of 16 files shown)


appeals/api/src/server/endpoints/integrations/integrations.service.js
```diff
@@ -25,27 +28,56 @@ export const produceAppealUpdate = async (
...
+		return false;
+	}
+};
+
+export const produceServiceUsersUpdate = async (
+	/** @type {any[]} */ users, // TODO: data and document types schema (PINS data model)
+	/** @type {string} */ updateType
+) => {
+	if (users.length > 0) {
+		const topic = producers.boSvcUsr;
```
james-hill-kc, line 76: Very minor, but does this need to be abbreviated? Could it be `boServiceUser` instead?

appeals/api/src/server/infrastructure/topics.js
```diff
@@ -1,6 +1,7 @@
 export const producers = {
 	boCaseData: 'appeal-bo-case',
-	boDocument: 'appeal-bo-document'
+	boDocument: 'appeal-bo-document',
+	boSvcUsr: 'appeal-bo-svcusr'
```
james-hill-kc, line 4: As above, I think this abbreviation could potentially be confusing - could it be renamed to `boServiceUser: 'appeal-bo-serviceuser'`?

james-hill-kc (commented):
Looks good, just a few very minor comments from me

appeals/api/src/message-schemas/pins-root.schema.json
```diff
@@ -199,6 +167,78 @@
 					"examples": ["1021477"]
 				}
 			}
+		},
+		"ServiceUser": {
```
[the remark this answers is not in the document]
fabfab-kc: As these are definitions rather than actual schemas / properties, I thought I capitalise them like classes... Not sure there is a convention but I suggest we keep it. There is already poor convention on some of the properties, but we need to stick to those as defined by ODW.

appeals/api/src/server/endpoints/integrations/integrations.service.js
fabfab-kc: yeah not a problem, it's just an internal label...

appeals/api/src/server/infrastructure/topics.js
fabfab-kc: I can rename the key, but the value is the actual service bus topic so need to stick to it.

PR #1679, opened 2023-10-17
Feature/boas 1292 unpublish

HadrianHughesKC:
## Describe your changes

* feat(api/applications): emit an Update event on Unpublish
This is to trigger the Azure function which will perform the unpublish.
* feat(functions): create a folder for a new unpublish function
* feat(functions): create an unpublish documents function
* feat(api/applications): set 'unpublishing' instead of 'unpublished'
Mimics the behaviour when publishing. We should set to `unpublishing`,
and then the Azure function should eventually set the document to
`unpublished`.
* feat(api/applications): create an endpoint to mark unpublished
* chore(api/applications): regenerate Swagger types
* feat(functions): make call to mark unpublished from Azure function
* chore(functions): move document-unpublish function into document-publish
Better to group these functions together since they're related.

[BOAS-1292](https://pins-ds.atlassian.net/browse/BOAS-1292)

## Type of change 🧩

- [x] Documentation update

[BOAS-1292]: https://pins-ds.atlassian.net/browse/BOAS-1292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/functions/document-unpublish/host.json
```diff
@@ -0,0 +1,15 @@
+{
```
chrisdalykinandcarta, line 1: So this isn't something we do much of right now, but we can probably group the unpublish-doocument and publish-document functions together if you're comfortable making this change? It makes sense that they'd live and change together, and they share some common code.

chrisdalykinandcarta: For context, just looking to make the amount of function apps we have as small as possible - since each new function app comes at a cost of a duplicated runtime and extra steps in the release pipeline. I think in future we might squash this number down even more, but do you think it makes sense for this change to do it right now?

chrisdalykinandcarta (approved):

PR #1705, opened 2023-10-23
Feature/boas 1085 invalid html

HadrianHughesKC:
## Describe your changes

* feat(functions): validate HTML src in document-check function
* test(functions): add tests for new HTML validation function
* feat(web/applications): add client side validation for HTML uploads
* fix(functions): re-fetch blob stream for validating HTML files
The checkMyBlob function will consume the data from the original blob
stream---it cannot be used twice.

Not able to test locally, but unit tests are passing.

[BOAS-1085](https://pins-ds.atlassian.net/browse/BOAS-1085)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1085]: https://pins-ds.atlassian.net/browse/BOAS-1085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/functions/document-check/check-document/__tests__/check-html.test.js
+++ b/apps/functions/document-check/check-document/__tests__/check-html.test.js
@@ -1,0 +1,39 @@
+import { validateHTML } from '../validate-html.js';
+
+describe('HTML validation', () => {
+	test('does not throw for a YouTube iframe', async () => {
+		const html = '<iframe src="https://www.youtube.com"></iframe>';
+		await expect(() => validateHTML(html)).not.toThrow();
+	});
+
+	test('throws an error for a non-YouTube src attribute', async () => {
+		const html = '<iframe src="https://www.google.com"></iframe>';
+		await expect(() => validateHTML(html)).rejects.toThrow();
+	});
+
+	test('throws an error if a script tag is encountered', async () => {
+		const html = `
+      <html>
+        <script>console.log('hello world')</script>
+        <iframe src="https://www.youtube.com"></iframe>
+      </html>
+    `;
+
+		await expect(() => validateHTML(html)).rejects.toThrow();
+	});
+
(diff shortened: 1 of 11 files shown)


apps/functions/document-check/check-document/__tests__/check-html.test.js
```diff
@@ -0,0 +1,39 @@
+import { validateHTML } from '../validate-html.js';
+
+describe('HTML validation', () => {
```
chrisdalykinandcarta, line 3: Would be good to validate this against an actual template we have on the live env (the YouTube embed template)

HadrianHughesKC: I've just tried testing it against [the real template](view-source:https://infrastructure.planninginspectorate.gov.uk/wp-content/ipc/uploads/projects/WW010003/WW010003-001323-ish2%20s1.html), but the HTML on this page contains a lot of comments which are stripped out by `sanitize-html` (there is no option to preserve them). So the validation fails.

I can't see a way around this with the `sanitize-html` library, and annoyingly I can't find a different library that would do what we need (verify that an HTML string matches a predefined template).

HadrianHughesKC: One option which is a round-about way but might work:

We store the template on the live env as a Handlebars/Nunjucks template with the iframe source parameterised, then
1. An HTML upload comes in
2. Use a regex to extract the iframe src
3. Render the stored template using the extracted src
4. Compare the rendered (correct) result with the actual (possibly invalid) upload HTML

chrisdalykinandcarta (dismissed):

PR #1709, opened 2023-10-24
fix(web): fix/BOAT-594-fix-add-remove-line

kandcdev:
## Describe your changes
https://pins-ds.atlassian.net/browse/BOAT-594

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/client/components/add-another/add-another.js
+++ b/appeals/web/src/client/components/add-another/add-another.js
@@ -51,8 +51,11 @@ function addRemoveButton(element) {
 	removeButton.classList.add(CLASSES.removeButton.additional);
 
 	const buttonContainer = element.querySelector(SELECTORS.removeButtonContainer) || element;
+	const buttonAlreadyExists = buttonContainer.querySelector('button');
 
-	buttonContainer.appendChild(removeButton);
+	if (!buttonAlreadyExists) {
+		buttonContainer.appendChild(removeButton);
+	}
 }
 
 function removeRemoveButtons(element) {
@@ -62,6 +65,7 @@ function removeRemoveButtons(element) {
 function bindItemEvents(componentInstance, item) {
 	item.querySelector(SELECTORS.removeButton).addEventListener('click', (event) => {
 		event.target?.closest(SELECTORS.item)?.remove();
+		initItems(componentInstance);
 		updateAddButtonState(componentInstance);
 	});
 }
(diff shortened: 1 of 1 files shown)


appeals/web/src/client/components/add-another/add-another.js
```diff
@@ -51,17 +51,21 @@ function addRemoveButton(element) {
...
-	buttonContainer.appendChild(removeButton);
+	if (!buttonAlreadyExists) {
+		buttonContainer.appendChild(removeButton);
+	}
 }
 
-function removeRemoveButtons(element) {
-	element.querySelectorAll(SELECTORS.removeButton).forEach((button) => button.remove());
+function removeRemoveButtons() {
+	document.querySelectorAll(SELECTORS.removeButton).forEach((button) => button.remove());
```
james-hill-kc, line 62: Why was the `element` parameter removed? This component is designed to support multiple instances, which is why most of the functions in this file only perform queries like this in the scope of a parent element or a component instance - this will query the whole document, which could cause problems on pages where there are multiple instances of the "add another" component

appeals/web/src/client/components/add-another/add-another.js
```diff
@@ -110,6 +114,11 @@ function addAnotherItem(componentInstance) {
 
 	items = getItems(componentInstance);
 
+	if (items.length > 1) {
+		addRemoveButton(items[0]);
```
james-hill-kc, line 118: Edit: sorry, I misread this as adding the remove button to the new item - I see what this is doing now

appeals/web/src/client/components/add-another/add-another.js
james-hill-kc: I think I see now, this is supposed to remove all remove buttons in the component - in that case you could just pass in `componentInstance` and change the `document.querySelectorAll` to `componentInstance.elements.root?.querySelectorAll`

appeals/web/src/client/components/add-another/add-another.js
james-hill-kc: Just a suggestion, but this might be clearer if it was moved into its own function (called something like `updateRemoveButtons`) and called from here

appeals/web/src/client/components/add-another/add-another.js
james-hill-kc: I've checked locally and this current code does cause a bug on pages with multiple instances of this component. For example, if you go to an appellant case, select "incomplete", check two checkboxes which show text fields, add an extra item to each of those, then click "remove" on one of them, it removes all the remove buttons from both sets of text fields

james-hill-kc (commented):
This is looking good overall, just a couple of comments and one suggestion from me

james-hill-kc (approved):
Looks good!

PR #1769, opened 2023-11-17
feat(web/applications): search and add project team members

lutamKC:
…ect team members (boas-1035)

## Describe your changes
1. I created a table on our DB containing userId (ie the azure reference string), caseId and role.
The role values are hardcoded in the frontend and no check is performed to make sure that the saved value is among the allowed options. This is similar to what we do in other parts of the application, but let me know if you think we need a more stable approach.
2. I created 3 endpoints to retrieve/update the users records in the DB. No particular issue in this, but I'm still learning to work on the backend, so maybe you'll find something to correct.
3. Here starts the tricky part. We basically just copied what appeals does. The FIRST TIME the user lands on a page displaying users info (name or email), a request to the Microsoft Graph API is sent. This request retrieves EVERY user in the AD groups (inspector, case team, officer) and store them in the cache. From this moment any time a user navigates to a page that displays users info, no request is sent and the users are simply retrieved from the cache.
4. The mechanism is: get role and userId from the internal DB => search in the cache for a user with that userId => merge the info stored in the cache with the role and case stored in the DB => display all.
5. The cache has an expire time of 2hours (easy to change) to prevent our list to be out of date in case users get added/modified
6. To send the first request an AD token is necessary. After that, to retrieve the cached users, the token check is skipped. This is what appeals does.
7. The search uses the cached users as well. When searching for a team member, you're supposed to have all the users info in the cache. If you dont', no big deal:  point 3 is automatically triggered. So hitting the button Search with the query "Ludov" will start a search inside the cached users. NO GRAPH REQUEST IS SENT. Specifically, the search is performed with a very basic cachedUser.name.includes("Ludov")`. If you know a more efficient way to do it, I'm all ears, but the point is that I removed the previous mechanism (ie sending a graph request with the $search parameter)

## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/applications/case/project-team/applications-project-team.service.js
```diff
@@ -9,29 +9,51 @@ import projectTeamADService from './application-project-team.azure-service.js';
...
  * @param {number} pageNumber
- * @returns {Promise<{results: PaginatedProjectTeamMembers, errors: ValidationErrors}>}
+ * @returns {Promise<{results: PaginatedProjectTeamMembers}>}
  */
-export const searchProjectTeamMembers = async (searchTerm, ADToken, pageNumber) => {
+export const searchProjectTeamMembers = async (searchTerm, allAzureUsers, pageNumber) => {
+	const searchResults = allAzureUsers.filter((azureUser) => {
+		const { givenName, surname, userPrincipalName: email } = azureUser;
+
+		const recordString = (
```
francescobrenci-AR, line 20: suggestion: use template literals - `${givenName} ${surname} ${email}`.includes(searchTerm)

apps/api/src/server/utils/mapping/map-project-team.js
```diff
@@ -0,0 +1,19 @@
...
+	return pick(teamMember, ['role', 'userId', 'createdAt']);
+};
+
+/**
+ *
+ * @param {import('@pins/applications.api').Schema.ProjectTeam[]} teamMembers
+ * @returns {{role: string, userId: string}[]}
+ */
+export const mapProjectTeamMembers = (teamMembers) => {
+	return teamMembers?.map((teamMember) => mapProjectTeamMember(teamMember));
```
HadrianHughesKC, line 18: This will return undefined if `teamMembers` is undefined, which doesn't match the return type. (Also may as well just pass the function directly to map rather than using a lambda.)

```suggestion
	return teamMembers?.map(mapProjectTeamMember) ?? [];
```

apps/web/src/server/applications/case/project-team/application-project-team.azure-service.js
```diff
@@ -56,16 +50,46 @@ export const searchADMember = async (searchTerm, ADToken) => {
...
 
-	// remodel response and filter out duplicate results (same user can be in multiple groups)
+	// remodel response and filter out null or duplicate results (same user can be in multiple groups)
 	return allResults
 		.map((c) => c.value)
 		.flat(1)
 		.filter(
-			(value, index, self) => index === self.findIndex((selfItem) => selfItem.id === value.id)
+			(value, index, self) =>
+				!!value && index === self.findIndex((selfItem) => selfItem.id === value.id)
```
HadrianHughesKC, line 59: **nit-pick/non-blocker:** better to be explicit when coercing the type.

```suggestion
				Boolean(value) && index === self.findIndex((selfItem) => selfItem.id === value.id)
```

apps/web/src/server/applications/case/project-team/applications-project-team.service.js
```diff
@@ -40,3 +62,58 @@ export const searchProjectTeamMembers = async (searchTerm, ADToken, pageNumber)
...
+	try {
+		const projectTeamMember = await get(`applications/${caseId}/project-team/${userId}`);
+
+		response = { projectTeamMember };
+	} catch (/** @type {*} */ error) {
+		pino.error(`[API] ${error?.response?.body?.error?.code || 'Unknown error'}`);
+
+		response = new Promise((resolve) => {
+			resolve({ errors: { query: 'An error occurred, please try again later' } });
+		});
```
HadrianHughesKC, line 85: `Promise.resolve` should do the same.

```suggestion
		response = Promise.resolve({ errors: { query: 'An error occurred, please try again later' } });
```

or maybe even just using the value, not sure if it needs to be a promise.

```suggestion
		response = { errors: { query: 'An error occurred, please try again later' } };
```

lutamKC: this syntax is a way to trick JSDocs.
For some reason the RequestHandler functions of Express in this project expect the errors to be in the format 
``import('@pins/express').ValidationErrors``: a very complicated interface very different from the `Record<string, 'ERROR MESSAGE'>` that you would expect.

I'm not sure how to fix it, so returning a new Promise.resolve just confuses the type checking and allows me to return a reasonable error response without typing errors. Setting the value as `any` would probably help, but I try not to do it.

In this case, I did some changes to allow the Record<string,string> and the ValidationErrors to co-exist, so I could use your suggestion. However this same code is used everywhere else in the frontend.

francescobrenci-AR (approved):

PR #1784, opened 2023-11-22
feat(web): feature/BOAT-538-site-visit-manage

kandcdev:
## Describe your changes
[BOAT-538](https://pins-ds.atlassian.net/browse/BOAT-538)

## Type of change 🧩

- [ ] Documentation update

[BOAT-538]: https://pins-ds.atlassian.net/browse/BOAT-538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appeals/web/src/server/appeals/appeal-details/site-visit/__tests__/site-visit.test.js
+++ b/appeals/web/src/server/appeals/appeal-details/site-visit/__tests__/site-visit.test.js
@@ -8,7 +8,6 @@ const { app, installMockApi, teardown } = createTestEnvironment();
 const request = supertest(app);
 const baseUrl = '/appeals-service/appeal-details';
 const siteVisitPath = '/site-visit';
-const scheduleVisitPath = '/schedule-visit';
 const visitScheduledPath = '/visit-scheduled';
 const setVisitTypePath = '/set-visit-type';
 
@@ -20,7 +19,7 @@ describe('site-visit', () => {
 
 	describe('GET /site-visit/schedule-visit', () => {
 		it('should render the schedule visit page', async () => {
-			const response = await request.get(`${baseUrl}/1${siteVisitPath}${scheduleVisitPath}`);
+			const response = await request.get(`${baseUrl}/1${siteVisitPath}/schedule-visit`);
 			const element = parseHtml(response.text);
 
 			expect(element.innerHTML).toMatchSnapshot();
@@ -40,7 +39,7 @@ describe('site-visit', () => {
 		});
 
 		it('should re-render the schedule visit page with the expected error messages if required fields are not populated', async () => {
-			const response = await request.post(`${baseUrl}/1${siteVisitPath}${scheduleVisitPath}`).send({
+			const response = await request.post(`${baseUrl}/1${siteVisitPath}/schedule-visit`).send({
 				'visit-date-day': '',
 				'visit-date-month': '',
 				'visit-date-year': '',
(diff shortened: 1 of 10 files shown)


appeals/web/src/server/appeals/appeal-details/site-visit/__tests__/site-visit.test.js
```diff
@@ -18,290 +17,314 @@ describe('site-visit', () => {
 	});
 	afterEach(teardown);
 
-	describe('GET /site-visit/schedule-visit', () => {
-		it('should render the schedule visit page', async () => {
-			const response = await request.get(`${baseUrl}/1${siteVisitPath}${scheduleVisitPath}`);
-			const element = parseHtml(response.text);
+	for (const testType of ['schedule', 'manage']) {
```
james-hill-kc, line 20: I can see why you've taken this approach, but in principle we should avoid having logic like this in unit tests. This type of logic increases the risk of bugs in the test code itself, which may either cause false positives, hide issues, etc. If this were application code I would agree that avoiding the duplication is a good idea, but the same doesn't apply to unit tests (it's more important for them to be bulletproof than concise). I think you should remove this and simply duplicate the tests as required. There's some more info here: https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests

appeals/web/src/server/appeals/appeal-details/site-visit/__tests__/site-visit.test.js
```diff
@@ -311,12 +334,16 @@ describe('site-visit', () => {
 			nock.cleanAll();
 		});
 
-		it('should render the visit scheduled confirmation page', async () => {
-			const response = await request.get(`${baseUrl}/1${siteVisitPath}${visitScheduledPath}`);
-			const element = parseHtml(response.text);
+		for (const confirmationPageTypeToRender of ['new', 'unchanged', 'visit-type', 'date-time']) {
+			it(`should render the visit scheduled/rescheduled confirmation page of pageType "${confirmationPageTypeToRender}"`, async () => {
```
james-hill-kc, line 338: As a side note to the point about not having logic in tests, interpolating `confirmationPageTypeToRender` into the test name makes the resulting snapshot a bit confusing (for example, when reading the snapshots, it wasn't immediately clear what `site-visit GET /site-visit/visit-scheduled/:confirmationPageTypeToRender should render the visit scheduled/rescheduled confirmation page of pageType "date-time"` meant). In my opinion it should be easy to figure out if a given snapshot is correct by reading it's name/title and looking at the associated rendered HTML, so in this case I'd suggest something like `should render the visit changed confirmation page with updated time and date`

appeals/web/src/server/appeals/appeal-details/site-visit/site-visit.service.js
```diff
@@ -59,8 +59,8 @@ export async function updateSiteVisit(
 			json: {
 				visitType,
 				...(visitDate && { visitDate }),
-				...(visitStartTime && { visitStartTime }),
-				...(visitEndTime && { visitEndTime })
+				...{ visitStartTime },
```
james-hill-kc, line 62: if there's no risk of these being null or undefined, you should just be able to add them as shorthand object properties instead of using spread (eg. `json: { visitType, ...(visitDate && { visitDate }), visitStartTime, visitEndTime }`)

james-hill-kc (commented):
Looks good, just a few comments from me

james-hill-kc (approved):
Looks good to me!

PR #1814, opened 2023-11-30
fix(api): return removed member on team member remove api (boas-1309)

lutamKC:
## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/project-team/project-team.controller.js
+++ b/apps/api/src/server/applications/project-team/project-team.controller.js
@@ -75,7 +75,7 @@ export const removeProjectTeamMember = async ({ params, body }, response) => {
 	try {
 		await projectTeamRepository.remove(userId, Number(id));
 
-		response.send({});
+		response.send(projectTeamMember);
 	} catch {
 		throw new BackOfficeAppError(`Error while removing user ${userId} for the case ${id}`, 500);
 	}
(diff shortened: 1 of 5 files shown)


DavidJWhiteDev (changes requested):
just to make things easier in the api docs page, if we create definitions of the payloads for update and remove, this will make the api docs much easier to use, eg:

in swagger.js add:

UserGuidPayload: {
			type: 'object',
			properties: {
				userId: {type: 'string', description: 'Azure AD User Id', example: '1234-abcd-6789'}
			}
		},
		UserRolePayload: {
			type: 'object',
			properties: {
				role: {type: 'string', description: 'Project user role', example: 'case_manager'}
			}
		},
		
and these are used in the routes:
**patch update:**
#swagger.parameters['role'] = {
				in: 'body',
				description: 'New role',
				required: true,
				schema: { $ref: '#/definitions/UserRolePayload' },
			}

**remove-member:**
#swagger.parameters['userId'] = {
				in: 'body',
				description: 'Id of the user being removed',
				required: true,
				schema: { $ref: '#/definitions/UserGuidPayload' },
			}

lutamKC:
> should the remove-member endpoint be a DELETE rather than a POST?

it's not really a DELETING, but rather a removal from the list of team members.
I used the same concept of the API that removes an S51 Advice item from the publishing queue (POST `applications/${id}/s51-advice/remove-queue-item`).

For the same reason we don't have to worry about stuff like soft/actual deleting, because the data (i.e. the user info) will keep existing

DavidJWhiteDev (approved):
all good and working lovely!

PR #1874, opened 2023-12-11
feat(functions): register-representation function

JamesP-KC:
## Describe your changes

ASB-2069

Inclues:
- New function to consume events on the register-representation topic. Takes the message, which contains information about a Representation submitted by the Applications Front Office, and adds it to the Back Office database.
- Minor amendment to the POST /representations endpoint to allowing passing the `type` property

## Type of change 🧩

- [ ] Documentation update

- [] I have made corresponding changes to the documentation

### DIFF
--- a/apps/api/src/message-schemas/commands/register-representation.d.ts
+++ b/apps/api/src/message-schemas/commands/register-representation.d.ts
@@ -1,0 +1,11 @@
+export interface RegisterRepresentation {
+	referenceId: string;
+	caseReference: string;
+	representationType?: string;
+	originalRepresentation: string;
+	representationFrom: string;
+	registerFor: string;
+	represented: object;
+	representative: object;
+	dateReceived?: string;
+}
--- a/apps/api/src/message-schemas/commands/register-representation.schema.json
+++ b/apps/api/src/message-schemas/commands/register-representation.schema.json
@@ -20,6 +20,10 @@
 			"type": "string",
 			"description": "The unique reference of the case"
 		},
+		"representationType": {
+			"type": "string",
+			"description": "Type of representation"
+		},
 		"originalRepresentation": {
 			"type": "string"
 		},
@@ -38,6 +42,9 @@
 		"representative": {
 			"$ref": "../events/interested-party.schema.json",
 			"description": "Person or organisation submitting representation in the case of Agent representationFrom"
+		},
+		"dateReceived": {
+			"type": "string"
 		}
 	}
 }
--- a/apps/functions/register-representation/register-representation-command/index.js
+++ b/apps/functions/register-representation/register-representation-command/index.js
@@ -1,0 +1,57 @@
+import api from './back-office-api-client.js';
+import { isEmpty, pick } from 'lodash-es';
+
+const mapContactDetails = (entity) => {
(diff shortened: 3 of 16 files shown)


apps/functions/register-representation/register-representation-command/index.js
```diff
@@ -0,0 +1,59 @@
...
+	};
+};
+
+/**
+ *
+ * @param {import('@azure/functions').Context} context
+ * @param {import('@pins/api/src/message-schemas/commands/register-representation').RegisterRepresentation} msg
+ */
+export default async function (context, msg) {
+	context.log(`Handle register-representation event: ${JSON.stringify(msg)}`);
```
Yakitrak, line 39: Should we only log the ID for PII reasons?

apps/functions/register-representation/register-representation-command/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+	const caseId = await api.getCaseID(msg.caseReference);
+	if (!caseId) {
+		throw new Error(`No case found with reference: ${msg.caseReference}`);
```
Yakitrak, line 43:
```suggestion
	const caseReference = await api.getCaseID(msg.caseReference);
	if (!caseReference) {
		throw new Error(`No case found with caseReference: ${msg.caseReference}`);
```

Yakitrak: Just a minor suggestion for clarify but feel free to ignore this.

apps/api/src/message-schemas/commands/register-representation.schema.json
```diff
@@ -20,6 +20,10 @@
 			"type": "string",
 			"description": "The unique reference of the case"
 		},
+		"representationType": {
```
Yakitrak, line 23: Needs to be added to `required` array the top if we want it to be required. If not then we need to make it optional in the interface type file.

apps/api/src/message-schemas/commands/register-representation.d.ts
```diff
@@ -0,0 +1,11 @@
+export interface RegisterRepresentation {
+	referenceId: string;
+	caseReference: string;
+	representationType: string;
+	originalRepresentation: string;
+	representationFrom: string;
+	registerFor: string;
+	represented: object;
+	representative: object;
+	dateReceived: string;
```
Yakitrak, line 10:
```suggestion
	dateReceived?: string;
```

Yakitrak (commented):
Not familiar with this codebase as much as BO so will leave approval to them but left a few minor comments.

apps/functions/register-representation/register-representation-command/index.js
JamesP-KC: Yep quite right, this was for debugging so no longer needed. Thanks!

apps/functions/register-representation/register-representation-command/index.js
JamesP-KC: Thanks. I've updated the error message but the variable name is accurate as-is, the function returns `caseId` (Int) rather than `caseReference`(String) that Front Office uses

Yakitrak (approved):

PR #1881, opened 2023-12-12
fix(api/applications): remove caseId param from representation.getById

JamesP-KC:
## Describe your changes

Fixes issue with `GET /applications/{id}/representations/{repId}` endpoint where passing `caseId` parameter caused an error to be thrown due to Prisma's `findUnique` function only allowing a single parameter to be passed in the `where` section. Passing `{ id: 1, caseId: 2 }` is invalid so I have simplified it to be `{ id: 1 }` - removing `caseId`, as it is not needed anyway.

To replicate:
- Call the  `GET /applications/{id}/representations/{repId}` endpoint
  - for example, on dev: `GET /applications/2133/representations/7707`
- An error is returned: `Argument where of type RepresentationWhereUniqueInput needs exactly one argument, but you provided id and case.\nUnknown arg `case` in where.case for type RepresentationWhereUniqueInput`

Root cause appears to be passing `caseId` and `id` to `prisma.representation.findUnique`. 

This PR removes `caseId` as it is not used anywhere and not needed when using `findUnique`. Used to work OK with `findMany` but now redundant since https://github.com/Planning-Inspectorate/back-office/pull/1864

## Type of change 🧩

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/representations/representaions.service.js
+++ b/apps/api/src/server/applications/application/representations/representaions.service.js
@@ -13,12 +13,11 @@ export const getCaseRepresentations = async (caseId, pagination, filterAndSort) 
 
 /**
  *
- * @param {number} caseId
  * @param {number} repId
  * @returns {Promise<Prisma.RepresentationSelect>}
  */
-export const getCaseRepresentation = async (caseId, repId) => {
-	return representationsRepository.getById(Number(repId), Number(caseId));
+export const getCaseRepresentation = async (repId) => {
+	return representationsRepository.getById(Number(repId));
 };
 
 /**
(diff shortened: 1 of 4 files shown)


apps/api/src/server/applications/application/representations/representaions.service.js
```diff
@@ -18,7 +18,7 @@ export const getCaseRepresentations = async (caseId, pagination, filterAndSort)
  * @returns {Promise<Prisma.RepresentationSelect>}
  */
 export const getCaseRepresentation = async (caseId, repId) => {
```
francescobrenci-AR, line 20: worth removing the param caseId entirely since it's not needed anymore

JamesP-KC: Thanks @francescobrenci-AR, good call. I've removed it

francescobrenci-AR (approved):

PR #1986, opened 2024-01-17
Feature/boas 1369 redis

HadrianHughesKC:
## Describe your changes

* chore(web/applications): re-enable commented Redis setup code
* refactor(web/applications): use implicit return and guard statements
Using guard statements means a function always handles possible cases
and then culminates in the happy path. You can always find the happy
path at the bottom of the function.
* feat(web/applications): persist MSAL cache in redis
* chore(tooling): add redis dump.rdb to .gitignore
* test(web/applications): initialise Redis iff conn string is provided
* test(web/applications): increase memory limit for tests
* refactor(web/applications): move Redis client code into a shared package
* refactor(web/applications): change to use new @pins/redis package

[BOAS-1369](https://pins-ds.atlassian.net/browse/BOAS-1369)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1369]: https://pins-ds.atlassian.net/browse/BOAS-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -19,7 +19,7 @@
     "rollup": "node ./scripts/rollup/bundle-js.js",
     "sass": "node ./scripts/rollup/compile-css.js",
     "start": "node ./src/server/server.js",
-    "test": "cross-env NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules npx jest",
+    "test": "cross-env NODE_ENV=test NODE_OPTIONS=\"--experimental-vm-modules --max-old-space-size=8192\" npx jest",
     "test:cov": "cross-env NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules npx jest --coverage",
     "tscheck": "npx tsc -p jsconfig.json --maxNodeModuleJsDepth 0",
     "watch:rollup": "chokidar \"src/client/**/*.js\" --silent -c \"npm run rollup\"",
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -33,6 +33,7 @@
     "@pins/applications": "^0.0.0",
     "@pins/blob-storage-client": "^0.0.0",
     "@pins/platform": "^0.0.0",
+    "@pins/redis": "^0.0.0",
     "@planning-inspectorate/address-lookup": "^1.0.6",
     "@toast-ui/editor": "^3.2.2",
     "accessible-autocomplete": "2.0.4",
--- a/apps/web/src/server/app/config/session.js
+++ b/apps/web/src/server/app/config/session.js
@@ -45,18 +45,17 @@ function createRedisStore(redisConnectionString) {
  * @returns {import('express-session').Store}
  */
 function configureStore(redisConnectionString) {
-	if (redisConnectionString && typeof redisConnectionString === 'string') {
-		logger.info('Configuring Redis for session storage');
-		return createRedisStore(redisConnectionString);
+	if (!redisConnectionString) {
+		if (config.isProduction) {
+			throw new Error('REDIS_CONNECTION_STRING is required in production.');
+		}
+
+		logger.info('Configuring memory story for session storage');
+		return new session.MemoryStore();
 	}
-	// default to memory store if redis is not configured
-	// config.session.redis is required by config schema when NODE_ENV is production
-	// check here to be sure
-	if (config.isProduction) {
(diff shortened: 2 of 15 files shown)


apps/web/package.json
```diff
@@ -19,7 +19,7 @@
     "rollup": "node ./scripts/rollup/bundle-js.js",
     "sass": "node ./scripts/rollup/compile-css.js",
     "start": "node ./src/server/server.js",
-    "test": "cross-env NODE_ENV=test NODE_OPTIONS=--experimental-vm-modules npx jest",
+    "test": "cross-env NODE_ENV=test NODE_OPTIONS=\"--experimental-vm-modules --max-old-space-size=8192\" npx jest",
```
beejjacobs, line 22: As per Teams - this is OK for now as it's not directly caused by this PR, but let's investigate further and get this flag removed when we can.

apps/web/src/server/app/config/session.js
```diff
@@ -1,66 +1,65 @@
...
-// 		throw new Error('REDIS_CONNECTION_STRING is required in production');
-// 	}
-// 	logger.info('Configuring memory store for session storage');
-// 	return new session.MemoryStore();
-// }
+	logger.info('Configuring Redis for session storage');
+	return createRedisStore(redisConnectionString);
+}
 
 // TODO: for now we want to use memory store, until Redis setup is fixed
```
beejjacobs, line 61: Shall we remove this? :)

apps/web/src/server/lib/msal.js
```diff
@@ -1,6 +1,29 @@
 import msal, { LogLevel } from '@azure/msal-node';
 import config from '@pins/applications.web/environment/config.js';
 import pino from './logger.js';
+import * as redis from './redis.js';
+
+/**
+ * @param {*} cacheContext
```
beejjacobs, line 7: This can be 

```
 * @param {import('@azure/msal-node').TokenCacheContext} cacheContext
 ```

File: apps/web/src/server/lib/msal.js [hunk 4]
```diff
@@ -1,6 +1,29 @@
...
+const afterCacheAccess = async (cacheContext) => {
+	if (cacheContext.cacheHasChanged) {
+		await redis.set('msal', cacheContext.tokenCache.serialize());
+	}
+};
+
+const cachePlugin = {
+	beforeCacheAccess,
+	afterCacheAccess
+};
```
beejjacobs, line 26: We might want to move this to a shared package and make it a class, that takes a redis client. That way appeals can use it too.

Not required though if we want to check it works first.

apps/web/src/server/lib/redis.js
```diff
@@ -0,0 +1,21 @@
+import { init, get, set } from 'node-cache-redis';
```
beejjacobs, line 1: I think we should use the existing redis connection for session storage, and not create two connections to redis. We don't need this extra library, we've already got `redis` lib which gives us get and set functions. 

We should refactor so that we:
* parse conn string and create redis client (using existing lib)
* pass that client to session
* pass that client to msal plugin

beejjacobs (changes requested):
Looks like it'll work all OK - but we've got some duplication we should avoid. Happy to discuss!

HadrianHughesKC:
Thanks @beejjacobs I've made these updates now.

packages/redis/src/index.js
```diff
@@ -0,0 +1,44 @@
...
+		this.client.on('reconnecting', () => logger.info('Reconnecting to redis server...'));
+
+		// kick off the connection - no await here, in the background
+		this.client.connect().catch(onError);
+
+		// dev note: this may 'error' in vscode, but tscheck is all OK
+		this.store = new RedisStore({ client: this.client });
+
+    this.get = this.client.get;
+    this.set = this.client.set;
```
beejjacobs, line 40: Do we need these? store has the client, and cache plugin has the client? no harm them being their of course

beejjacobs (approved):
LGMT!

packages/redis/src/index.js
HadrianHughesKC: They're not needed, it's just a quality of life thing so you can do `redisClient.get()` instead of `redisClient.client.get()`.

PR #2008, opened 2024-01-22
Fix interaction between MSAL and Redis

HadrianHughesKC:
## Describe your changes

* refactor(packages/redis): use classes for MSAL plugin configuration
This is the way the documentation suggests to initialise an MSAL plugin.
Also, the MSAL library expects the Redis client to always return a
string, never null as is default in the redis npm package. The custom
MSALCacheClient class gets around this by returning an empty string if
the true result is null.
* refactor(web/applications): use guard structure in assertIsAuthenticated
* feat(packages/redis): use a DistributedCachePlugin with MSAL
This is required to use a custom cache like Redis.
* feat(web/applications): create new MSAL client when session updates
* chore(web/applications): remove now unnecessary MSALPlugin class
* fix(web/applications): require Redis when not running locally
* fix(web/applications): return a single global MSAL client for non-Redis
When not using Redis, the same MSAL client must be maintained in order
to keep track of session state.
* test(web/applications): add an optional flag to disable Redis
This is to allow unit tests to bypass Redis even though NODE_ENV=test

Tested locally and seems to solve the 'Too many redirects' issue.

[BOAS-1369](https://pins-ds.atlassian.net/browse/BOAS-1034)

## Type of change 🧩

- [ ] Documentation update

[BOAS-1369]: https://pins-ds.atlassian.net/browse/BOAS-1369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/web/.env.test
```diff
@@ -86,3 +86,5 @@ DOCUMENT_STORAGE_CONNECTION_STRING=AccountName=devstoreaccount1;AccountKey=Eby8v
 # The web hostname
 APP_HOSTNAME=localhost:8080
 AZURE_BLOB_STORE_HOST=any
+
+DISABLE_REDIS=true
```
francescobrenci-AR, line 90: do we need this on the .env file as well?

apps/web/src/server/app/auth/auth.guards.js
```diff
@@ -23,35 +23,43 @@ import * as authSession from './auth-session.service.js';
...
+			sessionAccount,
+			request.session.id
+		);
+
+		if (refreshedAuthenticationResult) {
+			pino.debug('Refreshed MSAL authentication.');
+			authSession.setAccount(request.session, refreshedAuthenticationResult);
+			return next();
 		}
+		// Destroy current session if refreshedAuthenticationResult not provided.
```
francescobrenci-AR, line 47: is this code still needed?

francescobrenci-AR (approved):

PR #2082, opened 2024-02-01
feat(web/applications): add retry option on got and blbostorage (boas…

lutamKC:
…-1472)

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/appeals/web/src/server/lib/graph-request.js
+++ b/appeals/web/src/server/lib/graph-request.js
@@ -3,15 +3,23 @@ import config from '@pins/appeals.web/environment/config.js';
 import got from 'got';
 import pino from './logger.js';
 
-const [requestLogger, responseLogger] = createHttpLoggerHooks(pino, config.logLevelStdOut);
+const [requestLogger, responseLogger, retryLogger] = createHttpLoggerHooks(
+	pino,
+	config.logLevelStdOut
+);
 
 export const prefixUrl = 'https://graph.microsoft.com/v1.0/';
 
 const instance = got.extend({
 	prefixUrl,
+	retry: {
+		limit: 3,
+		statusCodes: [500, 502, 503, 504]
+	},
 	responseType: 'json',
 	resolveBodyOnly: true,
 	hooks: {
+		beforeRetry: [retryLogger],
 		beforeRequest: [requestLogger],
 		afterResponse: [responseLogger]
 	}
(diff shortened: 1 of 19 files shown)


packages/blob-storage-client/src/blob-storage-client.js
```diff
@@ -1,6 +1,7 @@
 import { DefaultAzureCredential } from '@azure/identity';
```
lutamKC, line 1: this should cover every blob storage request. I don't know how to test it though.
For now, I suppose we could just test it on dev and make sure that it doesn't break anything.

appeals/web/src/server/lib/graph-request.js
```diff
@@ -3,15 +3,23 @@ import config from '@pins/appeals.web/environment/config.js';
...
+);
 
 export const prefixUrl = 'https://graph.microsoft.com/v1.0/';
 
 const instance = got.extend({
 	prefixUrl,
+	retry: {
+		limit: 3,
+		statusCodes: [500, 502, 503, 504]
+	},
```
lutamKC, line 18: this is it for what concerns GOT. A parameter in the extend method.
We could work for storing this value somewhere common and easy to change.
I'd love some advice, @beejjacobs

lutamKC: also turns out that, all over the code, we have many slightly different implementation of got.
We could, one day, refactor everything as a package.

@francescobrenci-AR @Maby0

fabfab-kc: This is a separate instance of got that is used to fetch Active Directory user info through MS Graph...

appeals/web/src/server/lib/middleware/add-apiclient-to-request.js
```diff
@@ -14,7 +17,12 @@ const instance = got.extend({
 
 const getInstance = (/** @type {string} */ userId) =>
 	instance.extend({
+		retry: {
```
fabfab-kc, line 20: This should be good for Appeals. It's a got instance injected in the request through middleware, so it can be configured per user.

francescobrenci-AR:
one thing you could do would be to abstract the retry config:
`{
  limit: 3,
  statusCodes: [500, 502, 503, 504]
}`

if we need to change any of it we don't need to go through each instance

beejjacobs (commented):
Looks sensible!

I think we should add config variables (env vars) for the retry limit + status codes, with the defaults you have specified (which look good to me). Then we don't have these values spread about the code base.

I do think a (separate) task around consolidating our api-clients is sensible, the web + functions could share a configurable api client.

BenSurgisonPINS (approved):
Looks ok to me

PR #2142, opened 2024-02-14
feat(applications): add event broadcasts nsip-project case team (BOAS-1434)

BenSurgisonPINS:
## Describe your changes

- Added a relation between Case and ProjectTeam in the prisma schema.
- Added an option to retrieve the related project team when getting the case data from the db.
- Amended building the nsip-project event data to include the project team data when available.
- Updated the api unit tests to support the changes.
- Added the ability to simulate the AD user data with dummy user data to make it easier to develop locally where AD is unavailable.

I have also tested the functionality above manually.
Reseeding has been tested to make sure it is unaffected.

## BOAS-1434 Add Event Broadcasts - nsip-project case team
https://pins-ds.atlassian.net/browse/BOAS-1434

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -132,3 +132,23 @@ AZURE_BLOB_STORE_ACCOUNT_KEY=
 FEATURE_FLAG_BOAS_10_TEST_FEATURE= "true"
 
 
+
+## Development #################################################################
+
+
+# Optional location of a json file containing dummy AD user data when AUTH_DISABLED=true.
+#
+# An example of one user in the file is as follows:
+# [
+#   {
+#     "id": "8ada869a-d397-4656-bd99-bea186e0ba6e",
+#     "surname": "Doe",
+#     "givenName": "John",
+#     "userPrincipalName": "john.doe@example.com"
+#   }
+# ]
+#
+# Make sure to create the json file in a temp folder (as below in the example) as it will not be checked in with your branch.
+#
+DUMMY_USER_DATA="/temp/dummy_user_data.json"
+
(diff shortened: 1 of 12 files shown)


apps/web/.env.example
```diff
@@ -132,3 +132,19 @@ AZURE_BLOB_STORE_ACCOUNT_KEY=
...
 
+
+## Development #################################################################
+
+
+# Optional location of a json file containing dummy AD user data when AUTH_DISABLED=true
+# An example of one user in the file is as follows:
+# [
+#   {
+#     "id": "1",
```
DavidJWhiteDev, line 143: can this be clear that it will be a guid, eg '1111-2222-3333'

DavidJWhiteDev (commented):
also check that re-seeding the DB works, as I know it clears down the case records etc in dev and Test, and we have had issues of the seeder failing after a schema change.

Otherwise all looks good, and I like the idea of the dummy data to enable us to easily work in local dev, nice one!

apps/web/.env.example
BenSurgisonPINS: I've changed the id to a guid in the example as suggested.

BenSurgisonPINS:
> also check that re-seeding the DB works, as I know it clears down the case records etc in dev and Test, and we have had issues of the seeder failing after a schema change.
> 
> Otherwise all looks good, and I like the idea of the dummy data to enable us to easily work in local dev, nice one!

Reseeding tested and works fine.

DavidJWhiteDev (approved):

PR #2343, opened 2024-04-03
Feature/boas 1585 add schema validation for all service bus events

BenSurgisonPINS:
## Describe your changes

Ensure that all service bus messages are validated against the schemas before broadcasting.

- Include all fields when sending an event message
- modifications to most DB repository functions to return full data
- Update unit tests to match

The scope of which actions produce an event broadcast is defined in the table in this ticket:
https://pins-ds.atlassian.net/browse/BOAS-1359

## BOAS-1585 Schema Validation for Service Bus
https://pins-ds.atlassian.net/browse/BOAS-1585

## BOAS-1655 Update code base after Service Bus Schema changes
https://pins-ds.atlassian.net/browse/BOAS-1655

## Type of change 🧩

- [ ] Documentation update


apps/api/src/server/utils/map-case-to-case-stage.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @param {string | null} stage
+ *
+ * @returns {import('pins-data-model').Schemas.Folder.caseStage.enum} caseStage
+ */
+export const folderCaseStageMapper = (stage) => {
+	let caseStage = stage?.toString().toLowerCase() ?? null;
```
HadrianHughesKC, line 7:
```suggestion
	const caseStage = stage?.toString().toLowerCase() ?? null;
```

HadrianHughesKC (dismissed):
Nice one 👍 I've left a few superficial suggestions, but it looks good to merge.

apps/api/src/server/utils/application-factory-for-tests.js
Maby0, line 1: could we separate this from the rest of the live code utils somehow? Maybe a test-utils folder somewhere?

apps/api/src/server/utils/map-case-to-case-stage.js
Maby0, line 1: Like the previous comment - could we separate this from the rest of the live code utils somehow? Maybe a test-utils folder somewhere?

BenSurgisonPINS: This is not just used for tests.  It is also used in: ` apps/api/src/server/infrastructure/payload-builders/folder.js`

Maby0: ah I see. I'd typically advocate against having tests use live util code since it becomes impossible to detect if that util code is wrong from the test run, but if it's just a switch-case function it's might be alright 😅

apps/api/src/server/utils/application-factory-for-tests.js
BenSurgisonPINS: I think we should do this in a separate ticket as it changes many files

HadrianHughesKC (approved):

PR #2346, opened 2024-04-04
fix(applications/web): project update width of content too narrow (BOAS-1587)

BenSurgisonPINS:
## Describe your changes

- Added styling to allow the content to span over the column reserved for the "Change" action link as requested.
- Tested manually including changing the width of the browser

## BOAS-1587 Project Update: Width of 'Content' too narrow
https://pins-ds.atlassian.net/browse/BOAS-1587

## Type of change 🧩

- [ ] Documentation update


HadrianHughesKC (changes requested):
It looks like one of the elements has the `govuk-grid-column-two-thirds` class, which is setting its width to `66.666....7%`. I think setting the width to anything over `100%` might cause bugs, so might be best to try removing this class.

DavidJWhiteDev:
are any tests / snapshots affected?

BenSurgisonPINS:
> It looks like one of the elements has the `govuk-grid-column-two-thirds` class, which is setting its width to `66.666....7%`. I think setting the width to anything over `100%` might cause bugs, so might be best to try removing this class.

I had a look at removing the two thirds but that resulted in the action column being to wide and an issue when narrowing the browser.  As each "dd" has the css "display: table-cell" I tried to apply the css property "column-span" and that didn't work either.  I really didn't want to go with the 150% solution but it worked and did so on the browsers I had available to me.  I would prefer to stick with that as it produces exactly what the ticket was asking for.

BenSurgisonPINS:
> are any tests / snapshots affected?

No, none are affected as the only change was to styling that is using a class that was already in the DOM.

HadrianHughesKC (dismissed):

HadrianHughesKC (approved):

PR #2355, opened 2024-04-08
feat(applications/web): allow developers to simulate address lookup (BOAS-1721)

BenSurgisonPINS:
## Describe your changes

These changes are also in part in preparation for allowing e2e tests to run locally as well as not requiring the developer to have the OS_PLACES_API_KEY set when running the application on localhost.

- Create wrapper around the address lookup function to allow simulation of address lookup
- Provide support for new DUMMY_ADDRESS_DATA environment variable
- Add documentation for how to use dummy data
- Tested manually 

## BOAS-1721 Allow developers to simulate address lookup
https://pins-ds.atlassian.net/browse/BOAS-1721

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/common/services/address.service.js
+++ b/apps/web/src/server/applications/common/services/address.service.js
@@ -14,12 +14,11 @@ import fs from 'node:fs/promises';
  * @returns {Promise<{addressList: address[]; errors?: AddressLookupValidationErrors}>}
  */
 export default async function (postcode, locals) {
-	if (config.dummyAddressData) {
-		const dummyAddressDataFile = path.join(process.cwd(), 'dummy_address_data.json');
-		const dummyAddressData = JSON.parse(await fs.readFile(dummyAddressDataFile, 'utf8'));
-		const matchingAddresses = { addressList: dummyAddressData[postcode] };
-		return Promise.resolve(matchingAddresses);
-	} else {
+	if (!config.dummyAddressData) {
 		return findAddressListByPostcode(postcode, locals);
 	}
+
+	const dummyAddressDataFile = path.join(process.cwd(), 'dummy_address_data.json');
+	const dummyAddressData = JSON.parse(await fs.readFile(dummyAddressDataFile, 'utf8'));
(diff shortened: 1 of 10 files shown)


apps/web/src/server/applications/common/services/address.service.js
```diff
@@ -0,0 +1,25 @@
...
+export default async function (postcode, locals) {
+	if (config.dummyAddressData) {
+		const dummyAddressDataFile = path.join(process.cwd(), 'dummy_address_data.json');
+		const dummyAddressData = JSON.parse(await fs.readFile(dummyAddressDataFile, 'utf8'));
+		const matchingAddresses = { addressList: dummyAddressData[postcode] };
+		return Promise.resolve(matchingAddresses);
+	} else {
+		return findAddressListByPostcode(postcode, locals);
+	}
+}
```
HadrianHughesKC, line 25: Just refactored to use a guard statement so the unhappy path is at the top out of the way.

Also since nothing is awaited, could the async be removed?

```suggestion
 * @returns {Promise<{addressList: address[]; errors?: AddressLookupValidationErrors}>}
 */
export default function (postcode, locals) {
	if (!config.dummyAddressData) {
		return findAddressListByPostcode(postcode, locals);
	}

	const dummyAddressDataFile = path.join(process.cwd(), 'dummy_address_data.json');
	const dummyAddressData = JSON.parse(await fs.readFile(dummyAddressDataFile, 'utf8'));
	const matchingAddresses = { addressList: dummyAddressData[postcode] };
	return matchingAddresses;
}
```

HadrianHughesKC (dismissed):
Nice one, looks good. I've left one small suggestion but going to approve. Let me know if you need a re-approval :)

apps/web/src/server/applications/common/services/address.service.js
BenSurgisonPINS: Change made as suggested except for the async as this is required due to the await on line 22 when reading the json file.

HadrianHughesKC (approved):

PR #2377, opened 2024-04-17
feat(functions): add azure function for general s51 advice migration

Maby0:
## Describe your changes
Adds capability to migrate general S51 advice from the curated db to the BO db.

Due to the high number of General S51 advices that need to be migrated (700-800), there is a paginated mechanism in place that returns (currently) 100 rows with each call. They are then sent to the api together to complete the migration before continuing with the rest of the advices.

This mechanism also allows for continuation of a migration process from the point it errored by noting the offset number and including it in the invocation request to the Azure Function.

Might not be necessary, but there is also a 'select' General S51 Advice migration function included in the same Azure Function, allowing for migration of a specific case.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/functions/applications-migration/common/constants.js
+++ b/apps/functions/applications-migration/common/constants.js
@@ -1,0 +1,2 @@
+export const ODW_GENERAL_CASE_REF = 'General';
+export const BO_GENERAL_CASE_REF = 'GS5110001';
--- a/apps/functions/applications-migration/common/constants.js
+++ b/apps/functions/applications-migration/common/constants.js
@@ -1,2 +1,0 @@
-export const ODW_GENERAL_CASE_REF = 'General';
-export const BO_GENERAL_CASE_REF = 'GS5110001';
--- a/apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
+++ b/apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
@@ -1,0 +1,53 @@
+import { QueryTypes } from 'sequelize';
+import { SynapseDB } from '../synapse-db.js';
+import { migrateS51AdviceForCase } from './s51-advice-migration.js';
+import { ODW_GENERAL_CASE_REF } from '../constants.js';
+
+/**
+ * @param {import('@azure/functions').Logger} logger
+ * @param {number} offset
+ */
+export const migrateGeneralS51Advice = async (logger, offset = 0) => {
+	const countQuery = `
+		SELECT COUNT(*) AS generalS51AdviceCount
+		FROM [odw_curated_db].[dbo].[nsip_s51_advice]
+		WHERE caseReference = '${ODW_GENERAL_CASE_REF}';
+	`;
+	const [{ generalS51AdviceCount }] = await SynapseDB.query(countQuery, {
+		type: QueryTypes.SELECT
+	});
+	logger.info(`Number of General S51 Advice found in odw_curated_db: ${generalS51AdviceCount}`);
+
+	const batchSize = 10;
+	while (offset < generalS51AdviceCount) {
+		const paginatedGeneralS51AdviceQuery = `
+			SELECT * FROM [odw_curated_db].[dbo].[nsip_s51_advice]
+			WHERE caseReference = '${ODW_GENERAL_CASE_REF}'
+			ORDER BY nsip_s51_advice.adviceId
+			OFFSET ${offset} ROWS
+			FETCH NEXT ${batchSize} ROWS ONLY;
+		`;
+
+		try {
+			await migrateS51AdviceForCase(logger, ODW_GENERAL_CASE_REF, paginatedGeneralS51AdviceQuery);
+		} catch (error) {
+			logger.error(`Batch with offset ${offset} failed to migrate.`, error?.response?.body, error);
+			throw error;
+		}
+		offset += batchSize;
+	}
+};
+
+/**
+ * @param {import('@azure/functions').Logger} logger
+ * @param {string[]} adviceIdList
+ */
+export const migrateSelectGeneralS51Advice = async (logger, adviceIdList) => {
+	for (const adviceId of adviceIdList) {
+		const query = `
+			SELECT * FROM [odw_curated_db].[dbo].[nsip_s51_advice]
+			WHERE adviceId = '${adviceId}'
+		`;
+		await migrateS51AdviceForCase(logger, ODW_GENERAL_CASE_REF, query);
+	}
+};
--- a/apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
+++ b/apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
@@ -1,7 +1,7 @@
 import { QueryTypes } from 'sequelize';
 import { SynapseDB } from '../synapse-db.js';
 import { migrateS51AdviceForCase } from './s51-advice-migration.js';
-import { ODW_GENERAL_CASE_REF } from '../constants.js';
+import { ODW_GENERAL_S51_CASE_REF } from '@pins/applications';
 
 /**
  * @param {import('@azure/functions').Logger} logger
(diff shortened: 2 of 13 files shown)


apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
```diff
@@ -0,0 +1,53 @@
...
+		SELECT COUNT(*) AS generalS51AdviceCount
+		FROM [odw_curated_db].[dbo].[nsip_s51_advice]
+		WHERE caseReference = '${ODW_GENERAL_CASE_REF}';
+	`;
+	const [{ generalS51AdviceCount }] = await SynapseDB.query(countQuery, {
+		type: QueryTypes.SELECT
+	});
+	logger.info(`Number of General S51 Advice found in odw_curated_db: ${generalS51AdviceCount}`);
+
+	const batchSize = 10;
```
JamesP-KC, line 21: We can probably increase the batch size, although I'm not sure what the upper limit should be. Did you notice any issues when trying to process all the records in one batch?

HadrianHughesKC: Also is it worth setting the batch size in some top-level config?

apps/functions/applications-migration/common/migrators/s51-advice-migration.js
```diff
@@ -42,21 +42,22 @@ export const migrateS51Advice = async (logger, caseReferences) => {
...
+		const [s51AdviceRows, count] = await SynapseDB.query(synapseQuery, {
 			replacements: [caseReference]
 		});
 
 		const s51AdviceEntities = s51AdviceRows.map((row) => {
-			removeValues(row, [null, 'None']);
-
+			if (caseReference === ODW_GENERAL_CASE_REF) {
+				mapGeneralToBoCaseRef(row);
+			}
```
HadrianHughesKC, line 56: Is the `mapGeneralToBoCaseRef` used elsewhere? If not I think it would be clearer to just inline it here.

```suggestion
			if (caseReference === ODW_GENERAL_CASE_REF) {
				row.adviceReference = row.adviceReference.replace(ODW_GENERAL_CASE_REF, BO_GENERAL_CASE_REF);
				row.caseReference = BO_GENERAL_CASE_REF;
			}
```

apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
Maby0: yeah, I agree, 10 is definitely too low. I'll try it locally with a large batch size to see if it causes issues. I put the batching in preemptively really so the api wouldn't have to deal with ~800 advices at once, but it might be alright

JamesP-KC: I think the main bottleneck with the API is the request body size. @HadrianHughesKC did we do some work to increase that a few months ago?

HadrianHughesKC: Yeah @JamesP-KC that rings a bell. I forget how far we got with it though. IIRC the default max request body is 30mb.

apps/functions/applications-migration/common/constants.js
```diff
@@ -0,0 +1,2 @@
+export const ODW_GENERAL_CASE_REF = 'General';
+export const BO_GENERAL_CASE_REF = 'GS5110001';
```
beejjacobs, line 2: Can we put these in the common `packages/applications` package please? (we should then use these in API + WEB, I think Francesco was going to make a ticket for that) - else this would be the third copy of this string!

apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
Maby0: Tested it locally and it does seem like it's able to do them all in one go (bear in mind that this will be ran separately to the typical case migration). Although, a low batchsize does allow you to restart the process at a certain point if something goes wrong with the migration of an advice. Not sure how useful this will be in practice though 🤔

apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
```diff
@@ -0,0 +1,53 @@
...
+		`;
+
+		try {
+			await migrateS51AdviceForCase(logger, ODW_GENERAL_CASE_REF, paginatedGeneralS51AdviceQuery);
+		} catch (error) {
+			logger.error(`Batch with offset ${offset} failed to migrate.`, error?.response?.body, error);
+			throw error;
+		}
+		offset += batchSize;
+	}
```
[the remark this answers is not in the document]
Maby0: the offset is a function parameter, as that way it lets you restart the process from a particular point if an advice migration fails mid-way through the process. Like I mentioned in my earlier comment though, not sure how useful the ability to do this will be in practice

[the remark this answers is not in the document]
HadrianHughesKC: Ah I see, that makes sense but because `offset` is a `number` type, it gets passed by value so mutating it won't persist once the function exits. If I'm understanding that right. I think you'd need to do something like return the offset that it reached from the function.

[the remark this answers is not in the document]
Maby0: aye, the idea is that when the function is invoked via a http post request (as in `apps/functions/applications-migration/general-s51-advice-migration/index.js`), you can optionally include the offset param (which you'll know the value of from the error logs if something's gone wrong) and it'll kick off the process again from the point of failure.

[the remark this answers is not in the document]
HadrianHughesKC: Ahh I see, so you get it from the error logs and set it manually? That sounds fine then 👍

apps/functions/applications-migration/common/constants.js
Maby0: I've created the addition to the `packages/applications`, but I haven't gone through the entire codebase replacing references to it yet. I started doing it and the # of file changes got a bit much, so will create a separate pr for it

beejjacobs: Yeah that's fine, let's do that separately.

apps/functions/applications-migration/common/migrators/s51-advice-migration.js
Maby0: done!

apps/functions/applications-migration/common/migrators/general-s51-advice-migration.js
Maby0: upped the batchsize to 100 for now

apps/functions/applications-migration/common/migrators/s51-advice-migration.js
```diff
@@ -81,9 +86,11 @@ export async function migrateS51AdviceForCase(log, caseReference) {
 const mapStatus = (status) =>
 	({
 		checked: 'checked',
```
DavidJWhiteDev, line 88: are these the values to be stored in our DB?  These look more like the schema values.  Our DB values are listed in the mapS51AdviceStatusToSchemaStatus fn where they are mapped to the schema values for schema payload.  But in this case should they not be: checked, not_checked, ready_to_publish, published, or donotpublish ?

DavidJWhiteDev (changes requested):

HadrianHughesKC (approved):

apps/functions/applications-migration/common/migrators/s51-advice-migration.js
JamesP-KC: If I remember correctly, this is to correct some values coming from ODW that do not conform to the schema. This mapping should be done in ODW's curated layer, which is covered in a ticket on their board: https://pins-ds.atlassian.net/browse/ODW-1122.

JamesP-KC (dismissed):

DavidJWhiteDev (approved):

PR #2380, opened 2024-04-19
fix(api/applications): correct multiple valid case statuses (boas-1737)

DavidJWhiteDev:
## Describe your changes

Case Migration, if run multiple times for a case, can create multiple case status records that are marked as valid.  This confuses the main app, as it expects only 1 CaseStatus record to be valid for any case.  The issue occurs in cases in Azure Prod and Test.

This fix does:
- add code to the migrator to set all previous CaseStatus records for a case to valid=0, before adding the latest CaseStatus record.
- SQL migration script to set all old CaseStatus records that are incorrectly marked as valid=1 to be 0.

## Migrated cases have multiple valid case status records
https://pins-ds.atlassian.net/browse/BOAS-1737

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/migration/migrators/nsip-project-migrator.js
+++ b/apps/api/src/server/migration/migrators/nsip-project-migrator.js
@@ -28,6 +28,7 @@ export const migrateNsipProjects = async (models) => {
 		await migrateCase(model);
 		await migrateApplicationDetails(model);
 		await migrateRegions(model);
+		await invalidateCaseStatuses(model);
 		await migrateCaseStatus(model);
 		await migrateCasePublishedState(model);
 		await migrateGridReference(model);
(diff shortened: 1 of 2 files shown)


francescobrenci-AR (approved):

apps/api/src/server/migration/migrators/nsip-project-migrator.js
```diff
@@ -28,6 +28,7 @@ export const migrateNsipProjects = async (models) => {
 		await migrateCase(model);
 		await migrateApplicationDetails(model);
 		await migrateRegions(model);
+		await invalidateCaseStatuses(model);
```
JamesP-KC, line 31: you could call this function within the `migrateCaseStatus` function, so all the "case status" related DB stuff is within a single fn. But happy either way so approved 👍

JamesP-KC (approved):

PR #2406, opened 2024-05-03
fix(functions): fix array validation for nsip-project

JamesP-KC:
## Describe your changes

Update the `nsip-project-migration` function to fix a couple of issues with transforming the ODW data into a JSON object that conforms to the schema

- Transform string/null values into array (`nsipOfficerIds`, `nsipAdministrationOfficerIds`, etc)
  - e.g. id of `"12345"` becomes `["12345"]` and `null` becomes `[]`
- Remove some values for required fields which were hardcoded to `null` (`jRPeriodEndDate`, `dateIAPIDue`, etc). These are now being returned by ODW so no longer need to be hardcoded to pass the schema validation.

https://pins-ds.atlassian.net/browse/APPLICS-225

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/functions/applications-migration/common/migrators/nsip-project-migration.js
+++ b/apps/functions/applications-migration/common/migrators/nsip-project-migration.js
@@ -59,22 +59,13 @@ const getNsipProjects = async (log, caseReference) => {
 	log.info(`Retrieved projects ${JSON.stringify(projects)}`);
 
 	return projects.map((project) => ({
-		dateIAPIDue: null,
-		jRPeriodEndDate: null,
-		operationsLeadId: null,
-		operationsManagerId: null,
-		caseManagerId: null,
-		nsipOfficerIds: [],
-		nsipAdministrationOfficerIds: [],
-		leadInspectorId: null,
-		inspectorIds: [],
-		environmentalServicesOfficerId: null,
-		legalOfficerId: null,
-		applicantId: null,
-		migrationStatus: null,
-		dateOfReOpenRelevantRepresentationStart: null,
-		dateOfReOpenRelevantRepresentationClose: null,
 		...project,
-		regions: project.region ? [project.region] : []
+		nsipOfficerIds: valueToArray(project.nsipOfficerIds),
+		nsipAdministrationOfficerIds: valueToArray(project.nsipAdministrationOfficerIds),
+		inspectorIds: valueToArray(project.inspectorIds),
+		migrationStatus: Boolean(project.migrationStatus),
+		regions: valueToArray(project.region)
 	}));
 };
+
+const valueToArray = (value) => (value ? [value] : []);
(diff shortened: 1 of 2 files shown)


apps/functions/applications-migration/common/migrators/nsip-project-migration.js
```diff
@@ -59,22 +59,13 @@ const getNsipProjects = async (log, caseReference) => {
...
-		regions: project.region ? [project.region] : []
+		nsipOfficerIds: valueToArray(project.nsipOfficerIds),
+		nsipAdministrationOfficerIds: valueToArray(project.nsipAdministrationOfficerIds),
+		inspectorIds: valueToArray(project.inspectorIds),
+		migrationStatus: project.migrationStatus ? Boolean(project.migrationStatus) : null,
+		regions: valueToArray(project.region)
 	}));
 };
+
+const valueToArray = (value) => (value ? [value] : []);
```
beejjacobs, line 71: Is it worth checking if value is an array? If ODW were to fix this issue, and return an array, then this would break.

apps/functions/applications-migration/common/migrators/nsip-project-migration.js
```diff
@@ -59,22 +59,13 @@ const getNsipProjects = async (log, caseReference) => {
...
-		applicantId: null,
-		migrationStatus: null,
-		dateOfReOpenRelevantRepresentationStart: null,
-		dateOfReOpenRelevantRepresentationClose: null,
 		...project,
-		regions: project.region ? [project.region] : []
+		nsipOfficerIds: valueToArray(project.nsipOfficerIds),
+		nsipAdministrationOfficerIds: valueToArray(project.nsipAdministrationOfficerIds),
+		inspectorIds: valueToArray(project.inspectorIds),
+		migrationStatus: project.migrationStatus ? Boolean(project.migrationStatus) : null,
```
beejjacobs, line 66: This is going to only return `true` or `null`, but `false` is a valid value. We probably want to default to `false`, because `null` will just be for those projects that are created new in the back office.

beejjacobs (commented):
A couple of quick thoughts.

apps/functions/applications-migration/common/migrators/nsip-project-migration.js
JamesP-KC: We should only ever get a string or null for the values this function is being used for. The data is the output of the SQL query so don't think we'll get an array, just a single value per row

apps/functions/applications-migration/common/migrators/nsip-project-migration.js
JamesP-KC: Thanks, that actually simplifies the logic a bit. If we get null/undefined, then `Boolean` will return `false`.

apps/functions/applications-migration/common/migrators/nsip-project-migration.js
beejjacobs: Hmm OK, I guess this is the downside of using the curated layer instead of service bus/an API - we're not _quite_ dealing with schema-compliant formats.

beejjacobs (approved):

PR #2409, opened 2024-05-07
feat(applications/api): Add migrationStatus 

Yakitrak:
## Describe your changes

Tickets:
- https://pins-ds.atlassian.net/browse/APPLICS-152 - Add `migrationStatus` to `migrate-case` function and endpoints
- https://pins-ds.atlassian.net/browse/APPLICS-224 - Add validation error reporting to migration endpoints

Ticket has been tested via swagger because data from ODW is failing validation and therefore the whole migrate-case function cannot be tested e2e.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/migration/migrators/nsip-project-migrator.js
+++ b/apps/api/src/server/migration/migrators/nsip-project-migrator.js
@@ -10,6 +10,7 @@ import {
 import * as caseRepository from '#repositories/case.repository.js';
 import { EventType } from '@pins/event-client';
 import { broadcastNsipProjectEvent } from '#infrastructure/event-broadcasters.js';
+import logger from '#utils/logger.js';
 
 const keyDateNames = allKeyDateNames.filter(
 	(name) =>
@@ -17,12 +18,13 @@ const keyDateNames = allKeyDateNames.filter(
 );
 
 /**
+ *
  * Handle an HTTP trigger/request to run the migration
  *
  * @param {import('pins-data-model').Schemas.NSIPProject[]} models
  */
 export const migrateNsipProjects = async (models) => {
-	console.info(`Migrating ${models.length} models`);
+	logger.info(`Migrating ${models.length} models`);
 
 	for (const model of models) {
 		await migrateCase(model);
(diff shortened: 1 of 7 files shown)


apps/api/src/server/migration/migrators/nsip-project-migrator.js
```diff
@@ -54,20 +55,30 @@ export const migrateNsipProjects = async (models) => {
 
 		const eventType = model.publishStatus === 'published' ? EventType.Publish : EventType.Update;
 
-		await broadcastNsipProjectEvent(nsipProject, eventType);
+		if (model.migrationStatus) {
+			console.info(`Broadcasting event for case ${model.caseId}`);
```
beejjacobs, line 59: Ideally we use the pino logger rather than console - so we get proper JSON formatted logs.

apps/functions/applications-migration/migrate-case/index.js
beejjacobs, line 1: Not sure why the change here with the for...of loop? It was much clearer in the main function and avoided the extra nesting now in `migrateCase`. Also the name of `migrateCase` is now misleading, as it migrates multiple cases.

Can we revert this change please?

beejjacobs (commented):
Quite a neat little change in the end. A few little comments.

apps/functions/applications-migration/migrate-case/index.js
Yakitrak: I updated this so the debug file can easily call the exported name function and test multiple case references (as close to what the function would do) from the newly created `debug.js` file. I can rename more appropriately if thats helpful. Or would you still prefer to revert?

apps/api/src/server/migration/migrators/nsip-project-migrator.js
Yakitrak: https://github.com/Planning-Inspectorate/back-office/pull/2409/commits/7efff32722d2f5f9900fd199edc3a2a71c637552

apps/functions/applications-migration/migrate-case/index.js
beejjacobs: Ah I see, that makes sense.

I think let's stick with the pattern we have elsewhere then, with a function for multiple cases, that calls a function for a single case, e.g.

`migrateCases` -> log + for loop
`migrateCase` -> migration logic

Then the debug file can call `migrateCases` (though if you want to keep it close to the function logic, you could also call the default exported function).

Yakitrak: Ah that makes sense, I have updated to follow the pattern and to be fair checking one case at a time probably the norm so I've kept it the same. Thank you!

beejjacobs (approved):
LGTM!

JamesP-KC (approved):
Nice work 👍

JamesP-KC (approved):

PR #2433, opened 2024-05-14
Fix/applics 261 s51 transform data title

Maby0:
## Describe your changes
This transforms the format of the title field that comes in from ODW (typically in a format that we don't use) to a format that the FO will use to display to the enduser.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/migration/migrators/s51-advice-migrator.js
+++ b/apps/api/src/server/migration/migrators/s51-advice-migrator.js
@@ -85,6 +85,7 @@ const mapModelToS51AdviceEntity = async ({
 }) => {
 	let firstName, lastName;
 	if (from) {
+		from = removeUnnecessarySpaces(from);
 		let otherNames;
 		[firstName, ...otherNames] = from.split(' ');
 		lastName = otherNames.join(' ');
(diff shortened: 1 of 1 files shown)


apps/api/src/server/migration/migrators/s51-advice-migrator.js
```diff
@@ -132,3 +141,20 @@ const mapStatus = (status) =>
...
+	const adviceNameParams = { enquirer, firstName, lastName };
+	return enquiryMethod === 'meeting'
+		? `Meeting with ${generateAdviceName(adviceNameParams)}`
+		: `Advice to ${generateAdviceName(adviceNameParams)}`;
+};
+
+const generateAdviceName = ({ enquirer, firstName, lastName }) => {
+	if (enquirer) {
+		return enquirer;
+	} else if (firstName && lastName) {
```
Yakitrak, line 155: Do we want to return just firstName and lastName is isolation if only one of those exists?

Yakitrak (approved):

apps/api/src/server/migration/migrators/s51-advice-migrator.js
Maby0: good spot, addressed!

PR #2463, opened 2024-05-31
feat(functions): azure function to transform html files for specific …

Maby0:
…applicationId/Refs

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/functions/applications-migration/common/back-office-api-client.js
+++ b/apps/functions/applications-migration/common/back-office-api-client.js
@@ -5,7 +5,32 @@ import { addAuthHeadersForBackend } from '@pins/add-auth-headers-for-backend';
 const config = loadApiConfig();
 
 /**
- * Handle an HTTP trigger/request to run the migration
+ *
+ * @param {import('@azure/functions').Logger} logger
+ * @param {string} path
+ *
+ * @returns {Promise<any>}
+ */
+export const makeGetRequest = async (logger, path, params = {}) => {
+	const reqParams = new URLSearchParams();
+
+	Object.keys(params).forEach((key) => {
+		const value = params[key];
+		if (Array.isArray(value)) {
+			value.forEach((v) => reqParams.append(key, v));
+		} else {
+			reqParams.append(key, value);
+		}
+	});
+	const request = constructAuthenticatedRequest();
+	const requestUri = constructUri(path);
+
+	logger.info(`Making GET request to ${requestUri}`);
+	const response = await request.get(requestUri, { responseType: 'json' });
+	return response.body;
+};
+
+/**
  *
  * @param {import('@azure/functions').Logger} logger
  * @param {string} path
@@ -14,13 +39,18 @@ const config = loadApiConfig();
  * @returns {import('got').CancelableRequest<any>}
  */
 export const makePostRequest = (logger, path, body) => {
-	const scheme = config.apiHost.match(/localhost/) ? 'http' : 'https';
-	const requestUri = `${scheme}://${config.apiHost}${path}`;
+	const request = constructAuthenticatedRequest();
+	const requestUri = constructUri(path);
 
 	logger.info(`Making POST request to ${requestUri}`);
+	return request.post(requestUri, {
+		json: body
+	});
+};
 
+const constructAuthenticatedRequest = () => {
 	const serviceName = 'function';
-	const requestWithApiKey = got.extend({
+	const authenticatedRequest = got.extend({
 		hooks: {
 			beforeRequest: [
 				async (options) =>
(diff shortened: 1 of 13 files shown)


apps/functions/applications-migration/transform-html-files/index.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+export default async (context, { body: { applicationInfo } }) => {
+	try {
+		await transformMigratedHtmlFiles(context.log, applicationInfo);
+		context.res = {
+			status: 200,
+			body: { message: `Successfully transformed html files for applications ${applicationInfo}` }
+		};
+	} catch (error) {
+		console.error(`Failed to transform HTML files: ${error}`);
```
Yakitrak, line 23:
```suggestion
		context.error(`Failed to transform HTML files: ${error}`);
```

apps/functions/applications-migration/transform-html-files/transform-migrated-html-files.js
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @param {import('@azure/functions').Logger} logger
+ * @param {string | undefined} applicationRef
+ */
+const getApplicationId = async (logger, applicationRef) => {
+	const applicationObject = await makeGetRequest(
+		logger,
+		`/applications/reference/${applicationRef}`
+	);
+	return applicationObject.id;
```
Yakitrak, line 58: Do we need a null check in case it doesn't find anything?

apps/functions/applications-migration/transform-html-files/transform-migrated-html-files.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {import('@azure/functions').Logger} logger
+ * @param {number} applicationId
+ */
+const getHtmlDocumentVersions = async (logger, applicationId) => {
+	// need to modify to deal with paginated responses
+	const queryResponseData = await makeGetRequest(
+		logger,
+		`/applications/${applicationId}/documents?criteria=text/html`
+	);
+	return queryResponseData.items;
```
Yakitrak, line 72: Same as above for here?

apps/functions/applications-migration/common/back-office-api-client.js
```diff
@@ -56,27 +22,55 @@ export const makeGetRequest = async (logger, path, params) => {
...
-				}
+					})
 			]
 		}
 	});
+	return authenticatedRequest;
+};
 
-	const response = await requestWithApiKey.get(requestUri, { responseType: 'json' });
-	return response.body;
```
Yakitrak, line 81: Not sure if this change will break how makePostRequest is being used in the validation migration function. Your change seems good, but this returns a promise where as the previous change returned the body.

Yakitrak (changes requested):
Looks good overall, just a couple of minor comments. And 2 more general things to maybe consider:
1. Its great you have comments to mark code that can be removed after migration but would be good to have an identical comment everywhere so we can put a tech debt ticket that can search for that exact comment making it easy to find and remove.
2. As there are no HTML docs in ODW, this is untested right? Is it worth waiting for ODW to migrate 1 html doc and then testing this out locally?

apps/functions/applications-migration/common/back-office-api-client.js
Maby0: Yeah, I wanted to make the `makeGetRequest` match the `makePostRequest` function and return the raw response but didn't want to change something I wasn't familiar with and alter the the validate-migration function. So I've left it as is and it just returns the body of the response instead, which is what it was originally.

apps/functions/applications-migration/transform-html-files/transform-migrated-html-files.js
Maby0: fixed!

apps/functions/applications-migration/transform-html-files/transform-migrated-html-files.js
Maby0: fixed!

apps/functions/applications-migration/transform-html-files/index.js
Maby0: fixed!

Maby0:
> 2\. As there are no HTML docs in ODW, this is untested right? Is it worth waiting for ODW to migrate 1 html doc and then testing this out locally?

The code flow here doesn't interact with the curated layer. It runs under the assumption that the document-metadata migration and actual document migration (Aimii tool) has already been ran. It then queries our BO API for any HTML files that need to be reformatted and updates them in blob and in the document-version table accordingly. 
I was able to test this by generating mock data locally and changing a test html file in blob store.

Yakitrak (approved):
LGTM, nice work!

PR #2484, opened 2024-06-06
chore: msal-node update

Yakitrak:
## Describe your changes

Ticket: https://pins-ds.atlassian.net/browse/APPLICS-456

PR has included fixing lots of issues that arose from this update. FIrsty, besides `@azure/msal-node`, `@azure/msal-common` and `@azure/identity` are also updated to latest versions. 

`@azure/msal-node` does not export default anymore so we have to use `import * as msal from @azure/msal-node`. This means the mocking done for authentication does not work. On `apps/web/testing/app/mocks/msal.js`, we were overwriting the real value `msal.ConfidentialClientApplication = ConfidentialClientApplication` which is not recommended, as its not mocking but mutating the library. After updating to v2, this method doesn't work at all.

Simply doing a `jest.mock` does not work as expected either so fixing the tests will require further investigation. For now we will use `AUTH_DISABLED` to skip authentication for unit tests. 2 additional files will be skipped tests as well. We will need a new ticket for this investigating and refactoring to fix the tests.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/jest.config.js
+++ b/apps/web/jest.config.js
@@ -11,7 +11,7 @@ const require = createRequire(import.meta.url);
 const toastUiRequirePath = require.resolve('@toast-ui/editor');
 
 export default {
-	setupFiles: ['<rootDir>/setup-tests.js'],
+	setupFilesAfterEnv: ['<rootDir>/setup-tests.js'],
 	transform: {},
 	moduleNameMapper: {
 		'^uuid$': 'uuid',
--- a/apps/web/jest.config.js
+++ b/apps/web/jest.config.js
@@ -19,3 +19,8 @@ export default {
 	},
 	testTimeout: 30000
 };
+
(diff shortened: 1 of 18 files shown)


apps/web/jest.config.js
```diff
@@ -11,11 +11,16 @@ const require = createRequire(import.meta.url);
...
 		'^uuid$': 'uuid',
 		'^@toast-ui/editor$': toastUiRequirePath
 	},
 	testTimeout: 30000
 };
+
+process.env = {
+	...process.env,
+	AUTH_DISABLED: 'true'
+};
```
beejjacobs, line 26: I think this might be better in `setup-tests.js` or in `.env.test` file?

beejjacobs (approved):
LGTM!

PR #2566, opened 2024-07-22
Feature/applics 648 delete bespoke folders

Maby0:
## Describe your changes

Implement API functionality for deleting custom_folders.
This required introducing soft-delete for folders due to foreign key constraints against Documents when hard-deleting folders, so I've added a `deletedAt` property to the Folders schema. It's a date rather than a boolean due to the requirement of being able to recreate a folder with the same name after the initial one is deleted.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/file-folders/folders.service.js
+++ b/apps/api/src/server/applications/application/file-folders/folders.service.js
@@ -1,6 +1,7 @@
 import * as documentRepository from '#repositories/document.repository.js';
 import * as folderRepository from '#repositories/folder.repository.js';
 import { getPageCount, getSkipValue } from '#utils/database-pagination.js';
+import logger from '#utils/logger.js';
 import { mapDocumentVersionDetails } from '#utils/mapping/map-document-details.js';
 import {
 	mapBreadcrumbFolderDetails,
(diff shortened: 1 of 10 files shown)


apps/api/src/server/applications/application/file-folders/folders.service.js
```diff
@@ -113,6 +112,23 @@ export const getDocumentsInFolder = async (folderId, pageNumber = 1, pageSize =
...
+ * @param {Array<{ id: number, parentFolderId: number | null }>} folderList
+ * @returns {Promise<Array<{ id: number, parentFolderId: number | null }>>}
+ */
+export const getChildFolders = async (folderId, folderList = []) => {
+	const currentLevelFolderList = await folderRepository.getFoldersByParentId(folderId, {
+		select: { id: true, parentFolderId: true }
+	});
+	folderList.push(...currentLevelFolderList);
+
+	await Promise.all(currentLevelFolderList.map((folder) => getChildFolders(folder.id, folderList)));
```
HadrianHughesKC, line 128: Just thinking, `getChildFolders` returns a promise, so here you're creating an array of promises and then waiting for them all to resolve. But if say one of them fails, I think it will fail silently (it just won't push to `folderList`) which could be difficult to debug.

To get around it you could change `getChildFolders` so that rather than mutating its `folderList` argument, it just returns a new, updated list. Then you could use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) here instead of `Promise.all` to keep track of the list of promises. If any fail you'll get a result in the `rejected` state and you can throw an error or deal with the error some other way.

This should work as it is though, so this could be an improvement for later. Maybe worth leaving a tech debt note. Let me know what you think though.

HadrianHughesKC (dismissed):
One suggestion but could be done later. This looks fine 👍

apps/api/src/server/applications/application/file-folders/folders.service.js
Maby0: nice one! Addressed that in latest commit

HadrianHughesKC (approved):
lgtm

francescobrenci-AR (approved):

PR #2574, opened 2024-07-25
Fix/applics 743 description welsh

HadrianHughesKC:
## Describe your changes

* fix(web/applications): display Welsh description field if any English
* feat(web/applications): show Welsh name and description in check answers
feat(web/applications): always render Welsh fields if case is Welsh
* refactor(web/applications): make type coercions more explicit

[APPLICS-743](https://pins-ds.atlassian.net/browse/APPLICS-743)

## Type of change 🧩

- [ ] Documentation update

[APPLICS-743]: https://pins-ds.atlassian.net/browse/APPLICS-743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/web/src/server/applications/common/isCaseWelsh.js
+++ b/apps/web/src/server/applications/common/isCaseWelsh.js
@@ -1,4 +1,5 @@
 import { featureFlagClient } from '../../../common/feature-flags.js';
+import { getCase } from './services/case.service.js';
 
 /**
  * @param {Array<{name: string}>} regions
@@ -11,4 +12,21 @@ export const isCaseRegionWales = (regions) => {
 	}
 
 	return Boolean(regions.find((region) => region.name === 'wales'));
+};
+
+/**
+ * @param {number} caseId
+ * @returns {Promise<boolean>}
+ * */
+export const isCaseWelsh = async (caseId) => {
+	if (!featureFlagClient.isFeatureActive('applic-55-welsh-translation')) {
+		return false;
+	}
+
+	const caseObj = await getCase(caseId);
+	if (!caseObj.geographicalInformation) {
+		return false;
+	}
+
+	return isCaseRegionWales(caseObj.geographicalInformation.regions);
 };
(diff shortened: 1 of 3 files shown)


apps/web/src/server/applications/common/isCaseWelsh.js
```diff
@@ -12,3 +13,20 @@ export const isCaseRegionWales = (regions) => {
 
 	return Boolean(regions.find((region) => region.name === 'wales'));
 };
+
+/**
+ * @param {number} caseId
+ * @returns {Promise<boolean>}
+ * */
+export const isCaseWelsh = async (caseId) => {
```
DavidJWhiteDev, line 21: feel free to say no to this one Hadrian . . .
initially I was confused with the isCaseRegionWales and isCaseWelsh fn names.  I am wondering whether it is clear to change the isCaseWelsh name to something more reflective of "do we display the welsh fields"?  eg isCaseDisplayedWithWelshFields.  hmm, what are your thoughts?

DavidJWhiteDev (commented):
1 comment for consideration, otherwise lgtm!

apps/web/src/server/applications/common/isCaseWelsh.js
HadrianHughesKC: This function is general purpose rather than specific to the use case of checking whether to display the Welsh fields. It just check the regions to determine if one of them is Welsh, similar to the `isCaseRegionWales` function but you only need to pass it the case ID and it performs a fetch for you.

I could make the name for descriptive though, something like `isCaseInWales`. What do you think?

DavidJWhiteDev: hmm, that still sounds like the same name as isCaseRegionWales to me.  Maybe this is me overthinking it - No worries, proceed without changes.  I will approve.

DavidJWhiteDev (approved):

PR #2584, opened 2024-07-26
chore(tooling): delete specified cases (APPLICS-605)

BenSurgisonPINS:
## Describe your changes

A new entry script has been added to allow the ability to pass in a comma separated list of specified case numbers via the environment variable SPECIFIED_CASES to remove those specific cases and associated records as described in the ticket APPLICS-605.  

Please note that without the environment variable REMOVE_ALL_CASES being set to 'true', the transaction for each case is aborted to prevent the removal from actually taking place. 

I have successfully run this locally on my own machine against my local DB with the env var set to true and not set to true.  When set to true, the specified cases and all associated records are deleted successfully.

**Please note the environment variable SPECIFIED_CASES will only accept a comma separated list of case references that are prefixed "TRAIN" to prevent the deletion of any cases that are not training cases.**

## APPLICS-605 Remove Training cases from PROD (Specified cases)
https://pins-ds.atlassian.net/browse/APPLICS-605

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/packages/scripts/src/remove-cases.js
+++ b/packages/scripts/src/remove-cases.js
@@ -277,6 +277,9 @@ const removeCase = async (reference) => {
 		const startTime = Date.now();
 		await databaseConnector.$transaction(
 			async (tx) => {
+				if (!reference.startsWith('TRAIN')) {
+					throw new Error('Case is not a training case ' + reference);
+				}
 				const { id: caseId } = (await getCaseByRef(reference)) || {};
 				if (!caseId) {
 					throw new Error('No such case ' + reference);
@@ -335,7 +338,7 @@ const removeCases = async (references) => {
 	console.log(`Attempting to delete ${references.length} cases\n`);
 
 	for (const reference of references) {
-		const error = await removeCase(reference);
+		const error = await removeCase(reference.toUpperCase());
 		if (error) {
 			errors.push({ reference, error });
 		} else {
--- a/packages/scripts/src/remove-cases.js
+++ b/packages/scripts/src/remove-cases.js
@@ -10,6 +10,8 @@ const maxCasesToDelete =
 	parseInt(process.env.MAX_CASES_TO_DELETE ?? '0') || Number.MAX_SAFE_INTEGER;
 const maxWaitInSeconds = parseInt(process.env.TRANSACTION_MAX_WAIT_IN_SECONDS ?? '0') || 20;
 const maxTimeoutInSeconds = parseInt(process.env.TRANSACTION_TIMEOUT_IN_SECONDS ?? '0') || 100;
+const specifiedCases =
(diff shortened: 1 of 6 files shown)


packages/scripts/src/remove-cases.js
```diff
@@ -384,7 +389,20 @@ const getReferencesPrefixedWith = async (startsWith) => {
...
+ *
+ * @returns {Promise<void>}
+ */
+export const removeSpecifiedCases = async () => {
+	if (specifiedCases.length > 0) {
+		console.log('Specified cases to be removed:');
+		specifiedCases.forEach((reference) => console.log('- ' + reference));
+		console.log('');
+	}
+	await removeCases(specifiedCases);
```
HadrianHughesKC, line 407: Does `removeCases` do anything if the list `specifiedCases` list is empty? If not, might be best to just return early from this function when the list is empty, like:

```suggestion
	if (specifiedCases.length === 0) {
		return;
	}
	
	console.log('Specified cases to be removed:');
	specifiedCases.forEach((reference) => console.log('- ' + reference));
	console.log('');

	await removeCases(specifiedCases);
```

packages/scripts/src/remove-cases.js
```diff
@@ -275,6 +277,9 @@ const removeCase = async (reference) => {
 		const startTime = Date.now();
 		await databaseConnector.$transaction(
 			async (tx) => {
+				if (!reference.startsWith('TRAIN')) {
```
DavidJWhiteDev, line 280: I am wondering, if we should be filtering using the sector "Training", rather than filtering based on a string compare on the case reference?  As that is the sector for all Training cases.

packages/scripts/src/remove-cases.js
```diff
@@ -384,7 +389,20 @@ const getReferencesPrefixedWith = async (startsWith) => {
  * @returns {Promise<void>}
  */
 export const removeAllTrainingCases = async () => {
-	// const references = await getReferencesPrefixedWith('TRAIN0110004');
 	const references = await getReferencesPrefixedWith('TRAIN');
```
DavidJWhiteDev, line 392: likewise we could use the sector Training here too?

packages/scripts/src/remove-cases.js
BenSurgisonPINS: Discussed and resolved

packages/scripts/src/remove-cases.js
BenSurgisonPINS: Discussed and resolved

packages/scripts/src/remove-cases.js
BenSurgisonPINS: remove cases will report that there are 0 cases to remove and I thought that would be more useful.

DavidJWhiteDev (approved):
approved after discussion with Ben

PR #2588, opened 2024-07-31
fix(functions): convert s51 adviceid to number

JamesP-KC:
## Describe your changes

When running the s51 migrator function, the validation is failing because `adviceId` is not an integer. It is currently a string as that is how Sequelize stores `bigint` data from SQL Server (which is the type ODW are using for this column in their curated DB). Therefore, we need to convert it to Number/int before sending the data to the API.

## Type of change 🧩

- [ ] Documentation update


Yakitrak (commented):
Maybe worth removing caseId and adviceId from `s51AdviceProperties` array above for clarity?

JamesP-KC:
> Maybe worth removing caseId and adviceId from `s51AdviceProperties` array above for clarity?

How about `status` and `attachmentIds`, would it make sense to remove those too as they are being mapped/modified like caseId and adviceId?

Yakitrak (approved):

PR #2618, opened 2024-08-16
Fix/migration function clean up

Maby0:
## Type of change 🧩

- [ ] Documentation update


Yakitrak (changes requested):
Looks so much nicer, thanks for the tidy-up. One small thing, the handleMigrationResponse validation also needs to be updated for the single case reference:

image

Might require a bit of a refactor so it works for the single entity functions and the orchestrator.

apps/functions/applications-migration/common/handle-migration-with-response.js
```diff
@@ -3,23 +3,27 @@ import { validateMigration } from './validate-migration.js';
...
 	context,
 	caseReferences,
 	migrationFunction,
-	entityName
+	entityName,
+	allowCaseReferencesArray = false
 ) => {
-	if (!caseReferences || caseReferences.length === 0) {
+	const validationError = validateRequest(caseReferences, allowCaseReferencesArray);
+	if (validationError) {
```
Yakitrak, line 19: Wouldn't this always be true since we are returning an object

Yakitrak (changes requested):

apps/functions/applications-migration/common/handle-migration-with-response.js
Maby0: it'll be undefined if it doesn't fall into any of the `if` blocks

Yakitrak (approved):

PR #2663, opened 2024-09-18
feat(web): APPLICS-963 - Amend Training env header text

christopher-semmens:

https://pins-ds.atlassian.net/browse/APPLICS-963?atlOrigin=eyJpIjoiOTA5MWU4NDNhYjhmNGJiN2JlYWM4ZTkzYmE2NTBiYWMiLCJwIjoiaiJ9

## Type of change 🧩

- [ ] Documentation update


apps/web/src/server/lib/nunjucks-globals/get-web-env-config.js
```diff
@@ -0,0 +1,6 @@
+import config from '../../../../environment/config.js';
+
+export const webEnvConfig = {
+	env: config.env,
```
beejjacobs, line 4: FYI If I follow the infra and config through correctly this will be `production` for all Azure environments.

apps/web/src/styles/7-components/app-header.scss
```diff
@@ -12,6 +12,10 @@
   border-bottom: .5rem solid pins-colours.$pins-colours-green-200;
   margin-bottom: 1rem;
 
+	&[data-env="test"] {
```
francescobrenci-AR, line 15: we want to set the new background for the TRAINING env only

francescobrenci-AR (changes requested):

apps/web/src/server/lib/nunjucks-globals/get-web-env-config.js
francescobrenci-AR: also it's probably worth adding a `isTraining` check as we do for the other envs in `base-config.js`

christopher-semmens: Will environment.NODE_ENV be set to 'training' when on https://back-office-training.planninginspectorate.gov.uk/?

beejjacobs: Nope, `NODE_ENV` will always be `production`. We'll need to add a new env var for this if we need the environment name.

francescobrenci-AR: we could use APP_HOSTNAME instead of adding a new one? - it's unlikely to change

christopher-semmens: @francescobrenci-AR @beejjacobs It appears APP_HOSTNAME isn't present on all env and as mentioned above NODE_ENV is set to production for all Azure environments. So I've taken a bit of a different approach to determine the env and used hostname in a new middleware function that is applied to all routes. Let me know your thoughts on this approach.

francescobrenci-AR (approved):

DavidJWhiteDev (approved):

PR #2675, opened 2024-09-25
feat(web): manage key dates remove duplicate hint (APPLICS-872)

BenSurgisonPINS:
## Describe your changes

- Remove the duplicate hint
- Do not display hint if not set (note a blank hint appears to have a minimum of 3 characters so this is taken into account)
 
## APPLICS-927 CBOS- minor content update
https://pins-ds.atlassian.net/browse/APPLICS-927

## Type of change 🧩

- [ ] Documentation update


HadrianHughesKC:
Does the `.govuk-hint` div still need to be there if it's empty? I think semantically it would make sense to render it only on the condition that it'll have some text inside.

BenSurgisonPINS:
> Does the `.govuk-hint` div still need to be there if it's empty? I think semantically it would make sense to render it only on the condition that it'll have some text inside.

I've done this as suggested.

francescobrenci-AR (approved):

PR #2714, opened 2024-11-01
fix(api): folder mapping for filepaths with varying number prefixes a…

Maby0:
…ccounted for

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/migration/migrators/folder/folder-name-map.js
+++ b/apps/api/src/server/migration/migrators/folder/folder-name-map.js
@@ -235,7 +235,7 @@ const traversePartialMap = (
 	}
 	if (typeof currentLevelObject === 'string') {
 		cbosPathObject.cbosPathParts.push(currentLevelObject);
-		nextLevel ? cbosPathObject.failedToMatchParts.push(nextLevel) : null;
+		nextLevel && cbosPathObject.failedToMatchParts.push(nextLevel);
 		return cbosPathObject;
 	}
 
--- a/apps/api/src/server/migration/migrators/folder/folder-name-map.js
+++ b/apps/api/src/server/migration/migrators/folder/folder-name-map.js
@@ -1,199 +1,198 @@
 import { createFolders } from './folder.js';
 
 const partialMappingObject = {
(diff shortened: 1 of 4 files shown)


apps/api/src/server/migration/migrators/folder/folder-name-map.js
```diff
@@ -236,6 +235,7 @@ const traversePartialMap = (
 	}
 	if (typeof currentLevelObject === 'string') {
 		cbosPathObject.cbosPathParts.push(currentLevelObject);
+		nextLevel ? cbosPathObject.failedToMatchParts.push(nextLevel) : null;
```
Yakitrak, line 238: Curious on why you used a ternary here, it looks like it would work but the null at the end seems odd to me since we are not doing anything if nextLevel is falsy. If we are avoiding an if statement we could do `&&` and do `nextLevel && cbosPathObject.failedToMatchParts.push(nextLevel)` although I'd say if statement is more explicit. No need to change anything, just wondering.

Yakitrak (approved):

apps/api/src/server/migration/migrators/folder/folder-name-map.js
```diff
@@ -1,199 +1,198 @@
...
 			},
-			'02 - SoS Decision': 'SoS Decision'
+			'SoS Decision': 'SoS Decision'
 		}
 	},
-	'10 - Post Decision': {
+	'Post Decision': {
 		'Post-decision': {
-			'01 - Feedback': 'Feedback',
-			'02 - JR': 'Judicial review',
```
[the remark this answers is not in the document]
Yakitrak: Nevermind its not removing `JR`, on the PR diff it looks like its removed but its still there.

Yakitrak (approved):

apps/api/src/server/migration/migrators/folder/folder-name-map.js
Maby0: good spot, that would be nicer, have changed in recent commit

PR #2787, opened 2025-01-23
Fix/migration validation errors

JamesP-KC:
## Describe your changes

APPLICS-1256

A couple of fixes for errors that have been occurring when running the migration function:
- api: 404 response for get case endpoint not being returned correctly
- validate migration function: add optional call to fail gracefully in scenario where odw/bo data is missing a key

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/application.controller.js
+++ b/apps/api/src/server/applications/application/application.controller.js
@@ -95,7 +95,7 @@ export const queryApplications = async ({ params }, response) => {
 
 	const application = await getCaseByRef(String(params.reference));
 	if (!application) {
-		response.end(404);
+		response.status(404).end();
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


apps/api/src/server/applications/application/application.controller.js
```diff
@@ -95,7 +95,7 @@ export const queryApplications = async ({ params }, response) => {
 
 	const application = await getCaseByRef(String(params.reference));
 	if (!application) {
-		response.end(404);
+		response.status(404).end();
```
Yakitrak, line 98: This should be `.send` right?

JamesP-KC: I believe `res.status(404).end();` and `res.status(404).send();` do the same thing? I tried both and the response body and response headers are the same. 

But if we wanted to return a response body, we would need to do `res.status(404).send({ errors: `Application with reference '${params.reference} not found` });`

There's a mix of both elsewhere, but we do return an error most of the time:
https://github.com/search?q=repo%3APlanning-Inspectorate%2Fback-office%20.status(&type=code

Yakitrak (approved):

PR #2794, opened 2025-02-03
fix(api): correct rel rep agent publishing (applics-1290)

DavidJWhiteDev:
## Describe your changes

The representationFrom field in the service bus broadcast is currently being overridden if there is an agent, this is not correct, we should just be populating the value of the CBOS field Representation.representedType.

schema field representativeId - In the case where a rep was marked as having an agent, but later changed, we want to retain this information in CBOS in case they set the agent back again.  But we do not want to broadcast the agent id if they have not currently set the Representation.representedType field to be AGENT.  Therefore if Representation.representedTypeis AGENT, populate the schema field representativeId  with the agentId, else NULL.

Change the broadcast payload build to send NULL value for registerFor field.  This field is only used by Front Office when reps come into them from the public, not used CBOS side.  

- schema field representativeId to only have a value if Representation.representedTypeis AGENT, else NULL
- Also schema field registerFor should send a null value
- schema field representationFrom should just populate with the CBOS Representation.representedType value
- and update tests

## Relevant Representation publishing error - where represented person is an agent
https://pins-ds.atlassian.net/browse/APPLICS-1290

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/infrastructure/payload-builders/nsip-representation.js
+++ b/apps/api/src/server/infrastructure/payload-builders/nsip-representation.js
@@ -7,11 +7,12 @@
  */
 
 import { buildServiceUserPayload } from '#infrastructure/payload-builders/service-user.js';
+import { REPRESENTATION_FROM_TYPE } from '#api-constants';
 
 /**
  * Build Representation message event payload
  *
- * @param {function(RepresentationModel): Representation} representation
+ * @param {RepresentationWithFullDetails} representation
  * @returns {NSIPRepresentationSchema}
  */
 export const buildNsipRepresentationPayload = (representation) => {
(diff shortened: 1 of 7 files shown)


apps/api/src/server/infrastructure/payload-builders/nsip-representation.js
```diff
@@ -32,11 +33,12 @@ export const buildNsipRepresentationPayload = (representation) => {
...
 		representationType: mapRepresentationTypeToSchema(representation.type),
 		representedId: representation.represented?.id.toString() ?? null,
-		representativeId: representation.representative?.id.toString() ?? null,
-		representationFrom: representation.representative?.id
-			? 'AGENT'
-			: representation.representedType ?? null,
-		registerFor: representation.representedType ?? null,
+		representativeId:
+			representation.representedType == REPRESENTATION_FROM_TYPE.AGENT
+				? representation.representative?.id.toString()
```
Yakitrak, line 38:
```suggestion
				? representation.representative?.id?.toString()
```

DavidJWhiteDev: good call!

Yakitrak (approved):
LGTM

PR #2800, opened 2025-02-13
fix(api): fix db errors when adding new doc version (applics-1287)

DavidJWhiteDev:
## Describe your changes

when addind a new document version, the following fields are copied from the previous version, creating incorrect data - eg published blob paths etc:
- publishedBlobContainer 
- PublishedBlobPath 
- publishedStatusPrev 
- datePublished 
- redactedStatus 

These should all be reset to null.

## APPLICS-1287 CBOS DB data errors when adding new doc versions
https://pins-ds.atlassian.net/browse/APPLICS-1287

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/application/documents/document.service.js
+++ b/apps/api/src/server/applications/application/documents/document.service.js
@@ -381,6 +381,11 @@ export const createDocumentVersion = async (documentToUpload, caseId, documentId
 	currentDocumentVersion[0].size = documentToSendToDatabase.documentSize;
 	currentDocumentVersion[0].owner = documentToUpload.username;
 	currentDocumentVersion[0].originalFilename = documentToUpload.documentName;
+	currentDocumentVersion[0].datePublished = null;
+	currentDocumentVersion[0].publishedBlobPath = null;
+	currentDocumentVersion[0].publishedBlobContainer = null;
+	currentDocumentVersion[0].publishedStatusPrev = null;
+	currentDocumentVersion[0].redactedStatus = null;
 
 	await documentVersionRepository.upsert(currentDocumentVersion[0]);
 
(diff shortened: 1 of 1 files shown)


apps/api/src/server/applications/application/documents/document.service.js
```diff
@@ -381,6 +381,11 @@ export const createDocumentVersion = async (documentToUpload, caseId, documentId
 	currentDocumentVersion[0].size = documentToSendToDatabase.documentSize;
 	currentDocumentVersion[0].owner = documentToUpload.username;
 	currentDocumentVersion[0].originalFilename = documentToUpload.documentName;
+	currentDocumentVersion[0].datePublished = null;
+	currentDocumentVersion[0].publishedBlobPath = null;
+	currentDocumentVersion[0].publishedBlobContainer = null;
+	currentDocumentVersion[0].publishedStatusPrev = null;
+	currentDocumentVersion[0].redactedStatus = null;
```
Yakitrak, line 388: Question for my own understanding - wouldn't it be null anyway if we do not set a value as null here?


Also nitpick so feel free to ignore but should we change the const `currentDocumentVersion` to be the single version so we do not need to do `[0]` for all of these lines. Maybe I have missed on the process.

DavidJWhiteDev: no, because it was a copy of the previous version, fully populated with the data of that version.  Good question about the [0] though.

DavidJWhiteDev: @Yakitrak I have changed the array of 1 element to a single object

emil-placheta (approved):
lgtm

Yakitrak (approved):

PR #2807, opened 2025-02-20
Fix/applics 1305 breadcrumb error

JamesP-KC:
## Describe your changes

Fixes an error when building the breadcrumbs to show on the document publish success page. 

- This error only occurs when publishing Relevant Rep attachments, publishing other documents and S51 attachments is unaffected
- The cause is due to an expectation in a function (`getFolderIdFromFolderPath`) that the URL/path will be in a certain format. As the Relevant Representations page has a different path format to other Folders, this function fails (it returns `NaN`).

The fix:
- When visiting the Relevant Reps page, store the `/relevant-representations` page in session (`session.folderPage`) so we know where the user came from once they navigate to the attachment publish page.
- Add a condition so we look up the Relevant Reps `folderId` from the database. This is needed because the path to the Relevant Reps page doesn't contain a `folderId`, so we can't rely on `getFolderIdFromFolderPath`
- For belt and braces, also added an extra condition to `buildBreadcrumbItems` to guard against future scenarios where `folderId` is blank

APPLICS-1305

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/case/applications-case.locals.js
+++ b/apps/web/src/server/applications/case/applications-case.locals.js
@@ -108,11 +108,15 @@ export const registerDocumentGuid = async ({ params }, response, next) => {
  * @returns {Promise<BreadcrumbItem[]>}
  */
 export const buildBreadcrumbItems = async (caseId, folderId) => {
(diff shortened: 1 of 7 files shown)


emil-placheta (approved):
lgtm

PR #2868, opened 2025-06-26
fix(infrastructure): revert to DefaultRuleSet 1.0 in all envs but dev

beejjacobs:
Migration to FD Premium included the new ruleset, this caused some issues with E2E tests. Reverting for now, but keeping the new ruleset in DEV for testing.


DavidJWhiteDev (approved):
lgtm

PR #2882, opened 2025-07-14
fix(api): fix document reference not unique (applics-1082)

DavidJWhiteDev:
## Describe your changes

This ticket fixes several issues:
- when an S51 Attachment is added to the case, rather than finding the latest document reference number on a case in order to make the next one, it is counting the latest S51 Advice reference number.  So if a case has 10 docs numbered 00001 … 00010, but 3 S51 Advice, adding an S51 Advice attachment makes a duplicate 00004.

- The issue is further compounded by another fault, this time in the document side.  The fn that gets the latest document on a case uses created date order, rather than document reference.  So in the case above where the S51 attachment is made with 00004, this becomes the latest doc ref, and all documents added after that start from 00005 onwards, duplicating all existing docs.

- There is a 3rd issue in that deleted docs are ignored.  So if there are docs 00001 to 00004, and 00004 is deleted, the current code returns 00004 as the next number to use.  We should be including deleted docs in order to keep the doc ref unique. 

## APPLICS-1082 - Live bug: Document Reference numbers are not unique
https://pins-ds.atlassian.net/browse/APPLICS-1082

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/api/src/server/applications/s51advice/s51-advice.controller.js
+++ b/apps/api/src/server/applications/s51advice/s51-advice.controller.js
@@ -26,12 +26,14 @@ import * as caseRepository from '#repositories/case.repository.js';
 import {
 	makeDocumentReference,
 	createDocuments,
-	deleteDocument
+	deleteDocument,
+	getIndexFromReference
 } from './../application/documents/document.service.js';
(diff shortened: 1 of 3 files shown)


ypk (approved):
LGTM 👍🏻

PR #2893, opened 2025-07-25
Fix/applics 1596 handle special characters in document submissions

emil-placheta:
https://pins-ds.atlassian.net/browse/APPLICS-1596

This PR (and the [FO counterpart](https://github.com/Planning-Inspectorate/applications-service/pull/1483)) fix several issues that were occurring when files with special characters were being uploaded and affecting both applications. 

The issues  affecting CBOS:
- files containing special characters were stuck with 'awaiting virus check status' after being uploaded from FO
- files containing special characters uploaded directly to CBOS could be previewed, but not downloaded

To fix this, FO will no longer use filenames to generate azure blob URL. The changes in this PR modify the blob copying functionality to account for this change and provide a fallback to the old URL to maintain compatibility while the changes are being deployed.

When both applications generate and consume new URL, the fallback code introduced in this PR can be safely removed.

**NOTE: This change should be deployed first, before the changes from the FO PR in order to ensure the documents uploaded from FO will correctly make their way into CBOS**

## Type of change 🧩

- [ ] Documentation update


ypk (dismissed):

beejjacobs:
> Merging the newest changes, but omitting changes to redaction that are potentially causing issues with feature flags. I'll investigate after the HYS journey fixes are signed off for release

Apologies I just spotted this issue with feature flags just now! PR here to fix - https://github.com/Planning-Inspectorate/back-office/pull/2902

PR #2916, opened 2025-08-18
fix(web): add gds summary card borders to key dates display (applics-1548)

prashantproX:
## Describe your changes

Added borders around the key dates sections to make them look more professional and consistent with the GOV.UK design system. The changes just wrap the existing content in summary cards without changing how anything works. All the original functionality like editing dates and viewing information works exactly the same, but now each section has a clean border and the "Manage dates" button is properly positioned in the card header.

   APPLICS-1548       / https://pins-ds.atlassian.net/browse/APPLICS-1548 

## Type of change 🧩

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

- [X] I have performed a self-review of my own code
- [X] I have double checked this work does not include any hardcoded secrets or passwords
- [X] I have made corresponding changes to the documentation
- [X] I have provided details on how I have tested my code
- [X] I have referenced the ticket number above
- [X] I have provided a description of the ticket
- [X] I have included unit tests to cover any testable code changes


apps/web/src/server/views/applications/case-key-dates/key-dates-index.njk
```diff
@@ -1,6 +1,7 @@
 {% extends "../case/layouts/applications-case-layout.njk" %}
 {% from "govuk/components/accordion/macro.njk" import govukAccordion %}
 {% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
```
DeanH1990, line 4: Looks like this isnt actually used anywhere?

apps/web/src/server/views/applications/case-key-dates/key-dates-index.njk
```diff
@@ -19,23 +20,47 @@
...
+                      </a>
+                   </li>
+                </ul>
+             </div>
+             <div class="govuk-summary-card__content">
+                <div class='pins-key-dates-section'>
+                   {{createSectionTable(values) }}
+                </div>
+             </div>
+          </div>
```
DeanH1990, line 61: Im not sure if these summary cards are going to be used for other pages that display things other than dates but they could be made into their own higher level macro to avoid duplicating the boiler plate. Not sure if it would work when the first if condition uses another template though.

Not totally the same but there is a similar card macro in FO: https://github.com/Planning-Inspectorate/applications-service/blob/main/packages/forms-web-app/src/views/components/ui/chevron-card.njk these are used on the homepage and detailed information page to give an idea of the reusability

PR #2923, opened 2025-08-21
fix(web): make Screening and scoping its own group (applics-1553)

prashantproX:
## Describe your changes 
Refactored the Pre-application section in Key Dates management by splitting it into two logical groups for better user experience. The original single Pre-application section containing 9 date fields was divided into "Pre-application dates" (containing project initiation dates like first notification, website publication, submission dates, and Section 46 notification) and "Screening and scoping" (containing regulatory process dates for screening and scoping opinions). This change improves navigation and reduces cognitive load for case officers while maintaining full functionality - each section retains its individual "Manage dates" capability and all existing data storage/validation logic remains intact. The implementation uses a splitPreApplicationSection() function that transforms the data structure presentation-layer only, ensuring backward compatibility with the existing API.

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/case/key-dates/applications-key-dates.controller.js
+++ b/apps/web/src/server/applications/case/key-dates/applications-key-dates.controller.js
@@ -5,6 +5,73 @@ import {
 } from './applications-key-dates.utils.js';
 
 /**
+ * @param {Record<string, any>} sections - The original sections object
+ * @returns {Record<string, any>}  sections with grouped preApplication subsections
+ */
+function splitPreApplicationSection(sections) {
+	if (!sections?.preApplication) {
+		return sections;
+	}
+
+	const preApplicationData = sections.preApplication;
+
+	const preApplicationFields = [
+		'datePINSFirstNotifiedOfProject',
+		'dateProjectAppearsOnWebsite',
+		'submissionAtPublished',
+		'submissionAtInternal',
+		'section46Notification'
+	];
+
+	const screeningAndScopingFields = [
+		'screeningOpinionSought',
+		'screeningOpinionIssued',
+		'scopingOpinionSought',
+		'scopingOpinionIssued'
+	];
+
+	/** @type {Record<string, any>} */
+	const modifiedPreApplication = {};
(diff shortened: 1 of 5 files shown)


cprieststephens (changes requested):
Is there a way to add an additional 'Screening and scoping dates' box inside the 'Pre-application' accordion rather than adding a separate 'Screening and scoping dates' accordion?

apps/e2e/cypress/page_objects/keyDatesPage.js
```diff
@@ -5,7 +5,43 @@ export class KeyDatesPage extends Page {
 	elements = {};
 
 	clickManageDates(sectionName) {
```
cprieststephens, line 7: Do the Cypress commands in this function need to be chained using .then() to ensure that they run in the correct order?

apps/web/src/server/applications/case/key-dates/applications-key-dates.controller.js
```diff
@@ -4,6 +4,73 @@ import {
...
+			screeningAndScoping[field] = preApplicationData[field];
+		}
+	});
+
+	const result = Object.assign({}, sections);
+	delete result.preApplication;
+
+	return {
+		preApplication: {
+			preApplicationDates: modifiedPreApplication,
```
cprieststephens, line 55: Could preApplicationDates be renamed modifiedPreApplication for consistency? For example:

	return {
		preApplication: {
			modifiedPreApplication,
			screeningAndScoping
		},
		...result
	};

apps/web/src/server/views/applications/case-key-dates/key-dates-index.njk
```diff
@@ -13,10 +13,57 @@
 
 	{% for key, section in sections %}
 		{% set heading = key | keyDatesProperty %}
+
+		{% if key === 'preApplication' %}
+			{# Handle special case for preApplication with subsections #}
```
cprieststephens, line 18: Could a macro be created and used for both preApplication sections and other sections to avoid code duplication?

PR #3127, opened 2026-04-14
feat(api): added court decision outcome fields to post-decision key d…

ypk:
## Describe your changes

- key-dates.utils.js: added courtDecisionOutcome and courtDecisionOutcomeText to postDecisionDateNames allowlist
- map-key-dates.js: passed string fields through without Unix timestamp conversion
- key-dates.validators.js: excluded string fields from mapUnixTimestampToDate sanitiser
- post-decision-manage-dates.include.njk: added govukRadios with conditional govukInput for the Other option
- key-dates-property.js: added display labels for new fields
- applications-key-dates.validators.js: added validation requiring text when outcome is Other
- applications-key-dates.controller.js: preserved submitted values on validation failure, cleared courtDecisionOutcomeText when outcome is not Other
- key-dates-section.njk: added postDecision include branch
- key-dates.component.njk: added conditional rendering for Decision row

[[IDAS-59] Key Dates screen - Add new fields to Post-decision](https://pins-ds.atlassian.net/browse/IDAS-59)

## Type of change 🧩

- [x] Documentation update

[IDAS-59]: https://pins-ds.atlassian.net/browse/IDAS-59?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/api/src/server/utils/mapping/map-key-dates.js
+++ b/apps/api/src/server/utils/mapping/map-key-dates.js
@@ -1,3 +1,16 @@
+/**
+ * @file map-key-dates.js
+ * @description
+ * Utilities for mapping key date fields between API, DB, and event formats.
+ *
+ * Exports:
+ * - mapRequestToKeyDates: Flattens a sectioned request body into a DB-ready object, filtering by allowed keys.
+ * - mapKeyDatesToResponse: Maps DB record to sectioned API response (dates as Unix timestamps, string exceptions pass through).
+ * - mapKeyDatesToISOStrings: Maps DB record to flat object with all dates as ISO strings (for event broadcasting/schema compatibility).
+ *
+ * Most fields are Date objects; exceptions (see SECTION_STRING_FIELDS_EXCEPTIONS) are always strings.
+ */
+
 import { pick, mapValues } from 'lodash-es';
 import {
 	acceptanceDateNames,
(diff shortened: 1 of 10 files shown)


apps/api/src/server/utils/mapping/map-key-dates.js
```diff
@@ -86,7 +141,7 @@ export const mapKeyDatesToResponse = (keyDates) => {
 		examination: mapValues(pick(keyDates, examinationDateNames), mapDateToUnixTimestamp),
 		recommendation: mapValues(pick(keyDates, recommendationDateNames), mapDateToUnixTimestamp),
 		decision: mapValues(pick(keyDates, decisionDateNames), mapDateToUnixTimestamp),
-		postDecision: mapValues(pick(keyDates, postDecisionDateNames), mapDateToUnixTimestamp),
+		postDecision: mapValues(pick(keyDates, postDecisionDateNames), mapPostDecisionDates),
```
Tudor-PINS: Not doubting but just curious why we use a different mapping for the post decisions compared to everything else?

Tudor-PINS (commented):
LGTM, one question which isn't blocking approval. Please squash the two commits & let me know, then I can approve the PR.

apps/api/src/server/utils/mapping/map-key-dates.js
ypk: `postDecision` not only has dates, but also text strings with outcome details. Maybe a better way to resovle this is to rename the var completely

Tudor-PINS: Could we not just change the pick list to exclude the courtDecisionOutcome and courtDecisionOutcomeText and then run mapDateToUnixTimestamp? I don't see th eneed for making a new function if we're basiclaly just skipping those two things when we already have a pick() function too.

Tudor-PINS (dismissed):

apps/api/src/server/utils/mapping/map-key-dates.js
ypk: I've refactored the map functions to use a unified `mapDateField` function which resolves any issues?

Tudor-PINS (approved):

PR #3145, opened 2026-05-01
chore(web): move @planning-inspectorate/address-lookup into the codebase

beejjacobs:
@planning-inspectorate/address-lookup is from the legacy service-common repo, and only referenced by this project.

I have copied the address-lookup code as-is, switched it to ES6 imports, and replaced references to it.

Once merged this package will be removed/archived from service-common.


emil-placheta (approved):
nice one

PR #3155, opened 2026-05-14
feat/idas-221 - shapefiles upload processing and geojson generation

ypk:
## Describe your changes

When a ZIP file is uploaded to one of the "GIS Shapefiles" folder, after the virus scan completes the malware-detected function checks if the document belongs to the `GIS Shapefiles` folder by calling the existing properties and folder endpoints. If it does, it enqueues the document to the `shapefile-processing-queue` Service Bus queue.

The `process-shapefile` servicebus function picks it up, downloads the ZIP from blob storage, validates it to check if it has the required shapefile components (.shp, .shx, .dbf), then converts it to GeoJSON using `shpjs` library, uploads the GeoJSON alongside the source ZIP, and calls back to the API to create a new document version with the GeoJSON metadata.

If the ZIP is invalid it marks the document as invalid instead of retrying.

[\[IDAS-221\] Process shapefile uploads and generate GeoJSON](https://pins-ds.atlassian.net/browse/IDAS-221)

## Type of change 🧩

- [ ] Documentation update


apps/functions/applications-background-jobs/README.md
Tudor-PINS, line 1: I really don't think we need this here, especially in the codebase. If we're going to do architectural documentation, perhaps on Confluence, but I don't really think it should exist anyways.

apps/api/src/server/applications/applications.routes.js
```diff
@@ -76,7 +76,16 @@ router.patch(
...
+            schema: {
+                caseId: 1,
+                guid: 'a1b2c4d4-7ce5-410c-937e-28926dd7ab24',
+                status: 'awaiting_virus_check',
+                documentType: 'document',
+                documentURI: 'https://...',
+                caseRef: 'EN010001',
+                originalFilename: 'example.pdf',
+                dateCreated: '2026-05-08T22:55:11.000Z'
+            }
```
Tudor-PINS, line 84: Do we need / use all of these new fields on this API call?

apps/api/src/server/applications/application/__tests__/create-application.test.js
```diff
@@ -78,7 +78,9 @@ const expectedApplicantPayload = buildPayloadEventsForSchema(SERVICE_USER, {
 	webAddress: null
 })[0];
 
-jest.useFakeTimers({ doNotFake: ['performance'], now: 1_649_319_144_000 });
+// Keep real timers to avoid intermittent async stalls in full-suite runs.
+// This suite does not require timer control for its assertions.
+jest.useRealTimers();
```
Tudor-PINS: Did we run into issues with this causing the change?

apps/api/src/server/applications/application/documents/__tests__/create-document.test.js
Tudor-PINS, line 1: What is the thinking / need behind all the changes to this file?

apps/api/src/server/applications/application/documents/document.service.js
```diff
@@ -59,6 +61,37 @@ export const getFileNameWithoutSuffix = (documentNameWithExtension) => {
...
+		if (!doc.folderId) continue;
+		const folder = await getFolder(Number(doc.folderId));
+		if (folder?.displayNameEn === GIS_SHAPEFILES_FOLDER_NAME) {
+			doc.documentType = GIS_SHAPEFILE_DOCUMENT_TYPE;
+		}
+	}
+
+	return documents;
+};
+
```
emil-placheta: Why can't* the documentType be assigned during upload?

apps/api/src/server/applications/application/documents/shapefile-processing.controller.js
```diff
@@ -0,0 +1,59 @@
...
+		await markDocumentAsInvalid(documentGuid);
+		response.status(200).send({ status: 'invalid' });
+		return;
+	}
+
+	const { geoJsonFileName, geoJsonBlobPath, blobContainer, geoJsonSizeBytes } = body;
+
+	if (!geoJsonFileName || !geoJsonBlobPath || !blobContainer || geoJsonSizeBytes == null) {
+		throw new BackOfficeAppError(
+			'Missing required fields: geoJsonFileName, geoJsonBlobPath, blobContainer, geoJsonSizeBytes',
```
emil-placheta: Might be a bit confusing error, it's missing one of the required fields

apps/functions/applications-background-jobs/process-shapefile/src/apply-geojson-metadata.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Applies case metadata to a GeoJSON FeatureCollection as top-level properties.
+ *
+ * Note: `projectName` and `projectDescription` are intentionally omitted here.
+ * The nsip-document Service Bus event does not carry these fields. The API resolves
+ * the project name from the case record (Case.title) and populates it on the
+ * DocumentVersion at processing time.
```
emil-placheta: The comment is a bit misleading. We can just `The nsip-document Service Bus event does not carry these fields.` The rest of the comment relates to document version metadata, not GEOjson metadata that is being applied here.

apps/functions/applications-background-jobs/README.md
ypk: will move this to a confluence page

apps/api/src/server/applications/applications.routes.js
ypk: Yes, we need these because we hand off these details to process-shapefile which updates the document-properties fields

apps/functions/applications-background-jobs/process-shapefile/index.js
```diff
@@ -0,0 +1,160 @@
...
+
+		// Derive GeoJSON filename from the original ZIP filename
+		const baseFileName = (originalFilename ?? documentId).replace(/\.zip$/i, '');
+		const geoJsonFileName = `${baseFileName}.geojson`;
+
+		// Note: projectName is NOT available on the queue message.
+		// The API derives it from the case record (Case.title) using caseId.
+		const receivedDate = dateCreated ? new Date(dateCreated) : new Date();
+		const geoJsonWithMetadata = applyGeoJsonMetadata(geoJson, {
+			caseReference: caseRef ?? '',
```
ypk: this should never be empty

apps/api/src/server/applications/application/documents/shapefile-processing.controller.js
ypk: should say `missing one of the required fields`

apps/functions/applications-background-jobs/process-shapefile/src/apply-geojson-metadata.js
ypk: forgot to update this when I modularised the index

apps/api/src/server/applications/application/documents/shapefile-processing.controller.js
ypk: changed it to `Missing one or more required fields` as `one of the` mean one of X which would be misleading if more than 1 are missing

apps/functions/applications-background-jobs/process-shapefile/src/apply-geojson-metadata.js
ypk: done

apps/functions/applications-background-jobs/process-shapefile/index.js
ypk: added a check to ensure this will not be empty

apps/api/src/server/applications/application/__tests__/create-application.test.js
ypk: Yes, during development, jest.useFakeTimers was causing failures when running the tests. This is now reverted.

apps/api/src/server/applications/application/documents/__tests__/create-document.test.js
ypk: this was changed as the original implementation of `document.service.js` was changed, as it was reverted, this file is also reverted now.

apps/functions/applications-background-jobs/process-shapefile/index.js
```diff
@@ -0,0 +1,162 @@
...
+			documentURI
+		});
+		return;
+	}
+
+	try {
+		// Derive the private blob container and path from the documentURI
+		// documentURI format: <blobStorageUrl>/<container>/<path>
+		const blobName = extractBlobNameFromUri(documentURI);
+		const privateBlobContainer = config.BLOB_SOURCE_CONTAINER;
```
Tudor-PINS: Only the blob container name is derived from documentURI, not the path. Is this a bug or just an inaccurate comment?

apps/functions/applications-background-jobs/process-shapefile/src/validate-shapefile.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {{ files: Record<string, { dir?: boolean }> }} zipArchive
+ * @returns {string[]}
+ */
+export const getFileNamesFromZipArchive = (zipArchive) =>
+	Object.entries(zipArchive.files)
+		.filter(([, entry]) => !entry?.dir)
+		.map(([name]) => name.toLowerCase());
+
+/**
```
Tudor-PINS, line 16: what would happen here if people were to submit multiple shapefile names in one zip here? is there validation for this?

apps/api/src/server/applications/application/documents/shapefile-processing.service.js
```diff
@@ -0,0 +1,163 @@
...
+ */
+export const isDocumentInGisShapefilesFolder = async (documentGuid) => {
+	const document = await documentRepository.getById(documentGuid);
+	if (!document) {
+		logger.warn(`[SHAPEFILE] Document ${documentGuid} is not in GIS Shapefiles folder`);
+		return false;
+	}
+
+	const folder = await folderRepository.getById(document.folderId);
+	return folder?.displayNameEn === GIS_SHAPEFILES_FOLDER_NAME;
```
Tudor-PINS, line 33: can we match this on something other than the english display name? this seems fragile.

apps/functions/applications-background-jobs/README.md
```diff
@@ -0,0 +1,22 @@
+# Applications Background Jobs
```
Tudor-PINS: if we're moving the documentation to confluence, we can probably just delete this file entirely.

apps/web/src/server/views/applications/case-documentation/documentation-table.njk
```diff
@@ -88,11 +88,12 @@
 						{% if item.error %}
 							<span class='govuk-body-s govuk-!-font-weight-bold colour--red'>{{item.error}}</span>{% endif %}
 						{% set isPreviewActive = downloadNotDisabled and(item.mime === 'application/pdf' or item.mime === 'image/jpeg' or item.mime === 'image/png') %}
+						{% set displayFileName = item.originalFilename if item.originalFilename else item.fileName %}
```
Tudor-PINS, line 91: are the changes in this file needed? I don't see this change captured on the Jira ticket and it seems like separate work.

apps/api/src/server/applications/application/documents/document.service.js
ypk: We can set `documentType` at upload, but the existing `createDocuments` later calls `documentVersionRepository.upsert()` a second time with only blob storage fields after the file is stored. That second upsert overwrites `documentType` back to `null`.

apps/api/src/server/applications/application/documents/shapefile-processing.controller.js
```diff
@@ -0,0 +1,59 @@
...
+ *  2. Converted the shapefile to GeoJSON
+ *  3. Uploaded the GeoJSON to blob storage
+ *
+ * Body (success path):
+ *  { geoJsonFileName, geoJsonBlobPath, blobContainer, geoJsonSizeBytes }
+ *
+ * Body (failure path):
+ *  { invalid: true }
+ *
+ * @type {import('express').RequestHandler<{id: string; guid: string}, any, any, any>}
```
emil-placheta: Just to note, comments like these are unnecessary as they only describe what is happening rather than explain. 
Whe know where the controller is called, as we can easily follow the invocation/search for it when needed.

And the function is not complex/long enough to justify 20 lines of comments.

apps/api/src/server/applications/application/documents/shapefile-processing.service.js
```diff
@@ -0,0 +1,163 @@
...
+	const newVersionData = {
+		documentGuid,
+		version: newVersion,
+		fileName: geoJsonFileName,
+		originalFilename: geoJsonFileName,
+		description: caseRecord.description ?? '',
+		mime: 'application/geo+json',
+		size: geoJsonSizeBytes,
+		sourceSystem: 'back-office-applications',
+		stage: GIS_SHAPEFILE_STAGE,
```
emil-placheta, line 127: Why are we hardcoding the stage? The shapefile will be uploaded at two stages of the case, pre-app and acceptance if I remember right. That will allow us to see which files were uploaded as each stage, and look for potential differences. What is the reasoning behind hardcoding this, over inheriting the current case stage?

apps/api/src/server/applications/documents/documents.service.js
```diff
@@ -31,17 +59,39 @@ export const updateStatus = async (guid, status) => {
...
+		documentType: resolvedDocumentType,
+		documentURI: buildDocumentUri(
+			updatedDocument.privateBlobContainer,
+			updatedDocument.privateBlobPath
+		),
+		// @ts-ignore — getByDocumentGUID includes case.reference via a select; static type is incomplete
+		caseRef: document.case?.reference ?? null,
+		originalFilename: updatedDocument.originalFilename ?? null,
+		dateCreated: updatedDocument.dateCreated?.toISOString() ?? null
 	};
```
emil-placheta: I don't think this extra data being returned belongs here, the `updateStatus` is meant to update the status of the document, you would not expect it to return all this extra data. I feel like it makes the functionality a bit messy and forcing it into something that it is not meant to be. 

To achieve this, you also had to extend `getByDocumentGUID` to add a caseRef, which is also not what the funciton was initially designed to do, (which under the hood, performs a separate sql query anyway).  

If you need this data, simply query for it where you need it,  rather than bending current functionality in ways that will make it harder to maintain and intuitively understand

apps/api/src/server/repositories/document.repository.js
```diff
@@ -547,7 +547,12 @@ export const countDocumentsInFolder = (folderId) => {
 export const getByDocumentGUID = (documentGUID) => {
 	return databaseConnector.document.findUnique({
 		include: {
-			latestDocumentVersion: true
+			latestDocumentVersion: true,
+			case: {
+				select: {
+					reference: true
+				}
```
emil-placheta: This seems unnecessary.

Edit: addressed in document.service comment

apps/functions/applications-background-jobs/malware-detected/index.js
```diff
@@ -28,12 +42,33 @@ export const index = async (context, eventGridEvent) => {
...
-
 		eventData.scanResultType = isValidHtml ? eventData.scanResultType : 'Malicious';
 	}
 
-	const publishedStatusFromScanResult = getPublishedStatusFromScanResult(eventData.scanResultType);
-	await sendDocumentStateAction(guid, publishedStatusFromScanResult, context.log);
+	const publishedStatus = getPublishedStatusFromScanResult(eventData.scanResultType);
+
+	// Update document status in the DB; the enriched response tells us what kind of document it is
+	const documentInfo = await sendDocumentStateAction(guid, publishedStatus, context.log);
```
emil-placheta: (we could check for if the file belongs to GIS folder and assign document type here)

apps/web/src/server/views/applications/case-documentation/documentation-table.njk
ypk: This will show the complete file name with extension which was uploaded, rather than just the file name. This would be easy to identify what kind of file it is that failed, rather than hunting the properties from document history. Happy to split it into a separate PR if you'd prefer to keep this one scoped.

apps/functions/applications-background-jobs/process-shapefile/src/download-zip.js
```diff
@@ -0,0 +1,36 @@
...
+	for await (const chunk of downloadResponse.readableStreamBody) {
+		const buf = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk);
+		chunks.push(buf);
+	}
+
+	const buffer = Buffer.concat(chunks);
+	context.log(
+		`[SHAPEFILE] Downloaded ${blobName}: expected=${contentLength}, actual=${buffer.length} bytes`
+	);
+
```
emil-placheta: I don't think this is necessary and might pollute the logs with unnecessary information. Especially if you have a check for it 2 lines below.

apps/api/src/server/applications/application/documents/shapefile-processing.service.js
ypk: This is the same pattern used throughout the codebase e.g. `folderDocumentCaseStageMappings`, `isFolderApplicationDocuments` in `document.service.js`.

Folders don't have a stable type or slug field, The only identifier available us `displayNameEn`. 

I've extracted it into a constant `GIS_SHAPEFILES_FOLDER_NAME` so if the name ever changes, it's a single-place update.

emil-placheta (changes requested):
I like the separation and handling of the metadata separately for the GIS files, however, there are couple places where the functionaloty is unnecesarily complex and could be simplified. 
Ovreall looks good, just needs couple tweaks

apps/functions/applications-background-jobs/README.md
ypk: Removed. Though for the record, a README next to the code it documents doesn't go stale the way Confluence pages do. But sure, deleted.

apps/functions/applications-background-jobs/process-shapefile/src/validate-shapefile.js
ypk: That case is handled by `shpjs` library that we use; It returns an array of `FeatureCollections` for multi-layer ZIPs, and we flatten them into a single `FeatureCollection`. 

In case a multi-shapefile ZIPs, `shpjs` merges them if it deems valid.

apps/functions/applications-background-jobs/process-shapefile/index.js
ypk: The blob path is derived from `documentURI` via `extractBlobNameFromUri()` which strips the host and container segment and returns everything after.

The container comes from config `BLOB_SOURCE_CONTAINER`, not from the URI.

I've updated the misleading comment.

apps/api/src/server/applications/application/documents/shapefile-processing.controller.js
ypk: Fair point. I'll trim it to just the body shape docs.

apps/functions/applications-background-jobs/process-shapefile/src/download-zip.js
ypk: I'll remove the success log and keep only the warn for incomplete downloads.

apps/api/src/server/applications/documents/documents.service.js
ypk: Valid concern. The trade-off was: enrich one existing response vs adding a new endpoint + extra network call from the Azure Function.

I went with enrichment of existing response to avoid maintaining a separate endpoint.

The extra fields are backward-compatible (and existing consumers compleltely ignores them).

But if you'd prefer a separate endpoint for the routing info, happy to refactor this; that would mean one more API call from the function but a cleaner separation.

apps/api/src/server/repositories/document.repository.js
ypk: Agreed it's coupled to changes made in `documents.service.js`. If we move the routing data to a separate query, this include can be reverted.

apps/functions/applications-background-jobs/malware-detected/index.js
ypk: As this function doesn't have DB access, it can only know what the API tells it.

We'd still need the API to resolve the folder, which is what the enriched status response from `documents.service.js` does.

If we split it into a separate endpoint, the function would call that instead.

Either way the API does the folder lookup. We can just move the response to new API and get that data from there

apps/api/src/server/applications/application/documents/shapefile-processing.service.js
ypk: `stage: '0'` is per the requirement as it hides the GeoJSON from FO (same requirement as in S51/relevant reps).

But good shout on the stage validation: I've added a backend guard that rejects processing if the case isn't at pre-application or acceptance.

So even if someone bypasses the front-end, the API won't process it at other stages.

apps/api/src/server/applications/application/documents/shapefile-processing.service.js
Tudor-PINS: fair enough. I don't like it but if it's the pattern then it's the pattern!

apps/web/src/server/views/applications/case-documentation/documentation-table.njk
Tudor-PINS: @emil-placheta thoughts?

apps/api/src/server/applications/application/documents/document.routes.js
emil-placheta, line 1: thanks* for sorting this one

emil-placheta (dismissed):

emil-placheta (approved):

Tudor-PINS (approved):

apps/web/src/server/views/applications/case-documentation/documentation-table.njk
emil-placheta: I think that it's a simple quality of life improvement

PR #3156, opened 2026-05-15
Feat/idas 222 master geo json

DeanH1990:
## Describe your changes

adds a new api route to find all published gis boundary docs
when a document is published or unpublished all the published boundary docs are uploaded in a single master file geoJson which is saved straight to the published-documents blob so it can be accessed by FO

I have added tests and checked the functionality in the dev env, the master geoJson file successfully gets built in the published doc blob store

[ticket 222](https://pins-ds.atlassian.net/jira/software/c/projects/IDAS/boards/1034?selectedIssue=IDAS-222)

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/functions/applications-background-jobs/common/__tests__/master-geojson.test.js
+++ b/apps/functions/applications-background-jobs/common/__tests__/master-geojson.test.js
@@ -1,0 +1,216 @@
+import { Readable } from 'node:stream';
+import { jest } from '@jest/globals';
+
+const mockGetPublishedGisBoundaryDocuments = jest.fn();
+const mockUploadStream = jest.fn();
+const mockDownloadStream = jest.fn();
+
+jest.unstable_mockModule('../../gis-boundary-api-client.js', () => ({
+	getPublishedGisBoundaryDocuments: mockGetPublishedGisBoundaryDocuments
+}));
+
+jest.unstable_mockModule('../../blob-client.js', () => ({
+	blobClient: {
+		uploadStream: mockUploadStream,
+		downloadStream: mockDownloadStream
+	}
+}));
+
+jest.unstable_mockModule('../../config.js', () => ({
+	default: {
+		BLOB_PUBLISH_CONTAINER: 'published-documents'
+	}
+}));
+
+const { assertValidGeoJsonFeatureCollection, rebuildMasterGeoJson } = await import(
+	'../master-geojson.js'
+);
+
+describe('master-geojson', () => {
+	beforeEach(() => {
+		jest.resetAllMocks();
+	});
+
+	test('throws when GeoJSON is not a FeatureCollection', () => {
+		expect(() => assertValidGeoJsonFeatureCollection({ type: 'Feature' })).toThrow(
+			'Invalid GeoJSON FeatureCollection'
+		);
+	});
+
+	test('rebuilds the master GeoJSON from published GIS boundary documents', async () => {
+		// GIVEN
+		mockGetPublishedGisBoundaryDocuments.mockResolvedValue([
+			{
+				caseId: 1,
+				caseReference: 'EN0110001',
+				projectName: 'Project One',
+				documentGuid: 'D1234',
+				version: 2,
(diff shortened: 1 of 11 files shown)


apps/functions/applications-background-jobs/publish-document/index.js
```diff
@@ -89,4 +93,16 @@ export const index = async (
...
+		context.log(`Rebuilding master GeoJson after publishing GIS boundary ${documentId}`);
+
+		try {
+			await rebuildMasterGeoJson(context.log);
+		} catch (error) {
+			context.log(
+				`Failed to rebuild master GeoJson after publishing GIS boundary ${documentId}: ${error}`
+			);
+		}
+	}
```
Tudor-PINS, line 108: Is there an issue here with the fact that we are doing this after publishing? If this fails, all we do is log the error. Do we need to do more? I have the same question for this in the unpublish logic too.

apps/functions/applications-background-jobs/common/master-geojson.js
```diff
@@ -0,0 +1,104 @@
...
+
+const MASTER_GEOJSON_BLOB_PATH = 'gis-boundaries/all-project-boundaries.geojson';
+
+/**
+ * Streams all currently published GIS boundaries into one master GeoJSON blob.
+ *
+ * @param {import('@azure/functions').Logger} log
+ * @returns {Promise<void>}
+ */
+export const rebuildMasterGeoJson = async (log) => {
```
Tudor-PINS, line 59: Race condition in this function, will it ever be triggered?

apps/functions/applications-background-jobs/common/__tests__/master-geojson.test.js
Tudor-PINS, line 1: Could we maybe get more non happy-path tests here? Download / upload fails, and (if it's at all possible that we'd run into this in production) invalid JSON data?

DeanH1990: Have added a few sad path tests

apps/functions/applications-background-jobs/common/master-geojson.js
DeanH1990: Good spot. 

I asked Chris about this, his response was that it would be very unlikely for this to occur as it would be uncommon for more than one of these documents to published in a week

apps/functions/applications-background-jobs/publish-document/index.js
DeanH1990: My understanding is that by the time execution reaches this point, the publish/unpublish action itself has already succeeded and committed.

Because of that, I intentionally treated the master geoJson rebuild as a non blocking follow up action. If rebuilding the geoJson fails, we still want the publish/ unpublish operation to succeed rather than partially failing the user action after the document state has already changed.

In that scenario we currently log the failure so it can be investigated/ retried.

Happy to change the behaviour if we want the geoJson rebuild to become a hard dependency of publish/ unpublish.

apps/functions/applications-background-jobs/common/master-geojson.js
Tudor-PINS: Can we put a comment in the code acknowledging this decision maybe? Just in case we're ever debugging this in a few years time.

apps/functions/applications-background-jobs/publish-document/index.js
Tudor-PINS: OK, I get you. Have we considered using context.log.error on the errors to make them easier to find and fix issues if they do occur?

DeanH1990: updated

apps/functions/applications-background-jobs/common/master-geojson.js
DeanH1990: added

Tudor-PINS (approved):
lgtm just squash commits before pushing, ping me and I'll re-approve after checks have passed on squashed version

PR #3171, opened 2026-06-04
chore(deps): remove unused @planning-inspectorate/address-lookup package

beejjacobs:
Relates to #3145 - I hadn't removed the package from package.json and the lock file


emil-placheta (approved):
LGTM

PR #3177, opened 2026-06-11
fix(web): save rr notes when status changes (idas-285)

ypk:
## Describe your changes

- extracted latest referred action notes to prepopulate text field
- added a character limit to text field
- used govukCharacterCount component to display character counter
- added a fix to save referrals which are being wiped

[\[IDAS-285\] Save RR notes when user changes status](https://pins-ds.atlassian.net/browse/IDAS-285)

Currently every time a referral is made the notes are wiped, this means Case Officers and CAT team lose visibility of the full conversation history and cannot refer back to earlier notes. This fix changes the behaviour to outcome requested in ACs

## Type of change 🧩

- [ ] Documentation update

### DIFF
--- a/apps/web/src/server/applications/case/representations/representation-details/representation-status/representation-status-notes/representation-status-notes.view-model.js
+++ b/apps/web/src/server/applications/case/representations/representation-details/representation-status/representation-status-notes/representation-status-notes.view-model.js
@@ -2,6 +2,7 @@ import {
 	getRepresentationDetailsPageUrl,
 	getChangeStatusPageUrl
 } from '../representation-status.utils.js';
+import { NOTES_CHARACTER_LIMIT } from './representation-status-notes.constants.js';
 
 /**
  * @typedef {import('../../../relevant-representation.types.js').Representation} Representation
@@ -106,6 +107,11 @@ export const getRepresentationStatusNotesViewModel = (
 	representationDetails,
 	newStatus
 ) => {
+	const latestReferralAction = (representationDetails?.representationActions ?? []).find(
+		(action) => action.status === 'REFERRED'
+	);
+	const latestReferralNote = latestReferralAction?.notes ?? null;
+
 	return {
 		caseId,
 		repId,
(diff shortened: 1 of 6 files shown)


apps/web/src/server/views/applications/representations/representation-details/representation-status/representation-status-notes.njk
```diff
@@ -34,16 +34,21 @@
 						}
 					}) }}
 
-                    {{ govukTextarea({
+                    {{ govukCharacterCount({
                     name: "notes",
                     id: "notes",
+                    value: latestReferralNote,
```
cprieststephens, line 40: I found that if you enter an additional comment but don't select who the referral was made by from the radio options and the validation error message is displayed, the changes made in the comments box are lost. This may be because the latest referral note fetched from the DB is displayed when the page is re-rendered.

Do we need to display the data from the request body rather than the DB when the page re-renders due to errors?

cprieststephens (changes requested):

apps/web/src/server/views/applications/representations/representation-details/representation-status/representation-status-notes.njk
ypk: haha, I did notice that too, I've made Atilla and Alex aware of this issue during a call last week. This morning, the consensus is that we add an AC to fix this in this ticket rather than creating a new ticket to fix that

cprieststephens: Sounds good

apps/web/src/server/applications/case/representations/representation-details/representation-status/representation-status-notes/representation-status-notes.view-model.js
```diff
@@ -104,15 +105,23 @@ export const getRepresentationStatusNotesViewModel = (
...
 	repId,
 	representationDetails,
-	newStatus
+	newStatus,
+	submittedNotes = undefined
 ) => {
+	const latestReferralAction = (representationDetails?.representationActions ?? []).find(
+		(action) => action.status === 'REFERRED'
+	);
+	const latestReferralNote = submittedNotes ?? latestReferralAction?.notes ?? null;
```
cprieststephens: Should we fall back to `' '` here instead of `null` in case `null` is coerced into a string when rendered?

apps/web/src/server/applications/case/representations/representation-details/representation-status/representation-status-notes/representation-status-notes.view-model.js
```diff
@@ -104,15 +105,23 @@ export const getRepresentationStatusNotesViewModel = (
 	caseId,
 	repId,
 	representationDetails,
-	newStatus
+	newStatus,
+	submittedNotes = undefined
```
cprieststephens, line 110: Does submittedNotes need to be added to the above JSDoc comment with the other parameters?

ypk: good catch, I missed adding that in earlier. I've added it in.

